PR #81, opened 2022-11-07
Stores

tamitavares:
Create new stores that calling the controllers


KarinePistili (approved):
It looks great! We will check if it is all correct in the next milestone that will happen the first integrations with the new model structure :)

CamilaJar (approved):

PR #85, opened 2022-12-10
updated user sign-in authentication for firebase version 9, using the…

Na1au:
… new controllers


TalesAugusto1 (approved):
syntax and base controller changes, plus firebase upgrade to 9

PR #89, opened 2022-12-13
Stores

tamitavares:
Fix Constructors and Stores


package.json
```diff
@@ -13,7 +13,6 @@
     "bluebird": "^3.7.2",
     "chart.js": "^2.9.3",
     "core-js": "^3.6.4",
-    "csvtojson": "^2.0.10",
```
KarinePistili, line 16: What is this package for? Don't you need it anymore?

src/controllers/HeuristicController.js
```diff
@@ -1,101 +1,118 @@
...
-import HeuristicTest from "../models/HeuristicTest";
+import Heuristic from "@/models/Heuristic";
+import HeuristicAnswer from "@/models/HeuristicAnswer";
+import HeuristicQuestion from "@/models/HeuristicQuestion";
+import HeuristicQuestionAnswer from "@/models/HeuristicQuestionAnswer";
+import HeuristicQuestionDescription from "@/models/HeuristicQuestionDescription";
 
+// functions
 
+function toHeuristic(){
```
KarinePistili, line 11: Why are these functions outside the controller class? Is there an specific reason or it is because you had trouble calling it from another file?

src/controllers/TemplateController.js
```diff
@@ -1,76 +1,102 @@
...
-import {TemplateController} from '././TemplateController'
+import Template from "@/models/Template";
+import TemplateAuthor from "@/models/TemplateAuthor";
+import TemplateBody from "@/models/TemplateBody";
+import TemplateHeader from "@/models/TemplateHeader";
 
-let templateController = new TemplateController()
-*/
+// functions
+
```
KarinePistili, line 9: Same comment as the othe file, about the position of the functions

src/controllers/TestController.js
```diff
@@ -1,93 +1,121 @@
...
+import TestStructure from "@/models/TestStructure";
+import TestStructureOptions from "@/models/TestStructureOptions";
+import TestTemplateDoc from "@/models/TestTemplateDoc";
 
-// imports
+// functions
 
-import api from "@/api/index";
-import database from "../api/modules/database";
+function toTest(){
```
KarinePistili, line 11: Same thing, position of functions

src/controllers/UserController.js
```diff
@@ -1,91 +1,120 @@
...
+import User from "@/models/User";
+import UserAnswer from "@/models/UserAnswer";
+import UserTask from "@/models/UserTask";
+import UserTemplate from "@/models/UserTemplate";
+import UserTest from "@/models/UserTest";
 
-let userController = new UserController()
-*/
+// functions
+
```
KarinePistili, line 10: Position of functions?

src/views/admin/TestListView.vue
```diff
@@ -450,7 +450,7 @@ export default {
       this.subIndex = 0; //reset subIndex when main idex change
     },
   },
-  created() {
+  async created() {
     this.$store
```
KarinePistili, line 454: You forgot **await** before this dispatch. You created an async method so I suppose you want to make this dispatch asynchronous, so you need the await keyword or it won't work correclty

package.json
KarinePistili, line 16: What is this package for? Don't you need it anymore?

src/controllers/TestController.js
KarinePistili, line 11: Same thing, position of functions

src/controllers/TemplateController.js
KarinePistili, line 9: Same comment as the othe file, about the position of the functions

src/controllers/HeuristicController.js
KarinePistili, line 11: Why are these functions outside the controller class? Is there an specific reason or it is because you had trouble calling it from another file?

KarinePistili (changes requested):
I've left some comments on the files, small things to think about. Keep up the hard work!

PR #135, opened 2023-09-13
english error fix

TalesAugusto1:


jvJUCA (approved):
Correct english fix, and code format methods.

PR #217, opened 2023-10-09
when a test is deleted, all the cooperators subsequent tests will be …

TalesAugusto1:
…deleted as well now


jvJUCA (approved):
everything fine

PR #232, opened 2023-10-23
Feat/user test tab

Natanco1:
created a tab to display the user answers on the user tests


src/utils/statistics.js
KarinePistili, line 1: Please review @Natanco1

KarinePistili (changes requested):

KarinePistili (approved):

PR #246, opened 2023-11-14
reports users

JulioManoel:


jvJUCA (approved):
Succesfully showing reports in users

Issue #318, opened 2024-02-26
[🐞 BUG]:  default Portuguese translation for auth errors


Gaurav-Verma07:
### Description 📝

Default lang. as Portuguese in auth popup instead of selected lang.

### Link 🔗

/signin

### Steps to Reproduce 🔄

1. Go to login page.
2. Select `english` as default language.
3. Enter incorrect credentials and clink SIGN-IN.
4. A popup appears in portuguese lang.

### Screenshots 📸

![image](https://github.com/ruxailab/RUXAILAB/assets/63046396/ef5ebd6d-42b0-4846-a6fb-f03b9a5f78b9)

### Expected Behavior 🤔

The popup text should be in eng. lang or selected language.

### Actual Behavior 😱

Popup text is in Portuguese lang. irrespective of selected lang.

### Environment 🌍

- Vue Component- `src/views/public/SignInView.vue`

Gaurav-Verma07:
Hey @jvJUCA  I can resolve this

jvJUCA:
Ok, i will assign you, this is a i18n problem, the text in the alert is hardcoded in portuguese. :+1:

PR #319, opened 2024-02-26
fix: i18n translations for auth errors

Gaurav-Verma07:
Wrote auth error's i18n translations.

Related Issue: #318


jvJUCA (approved):
Ok, the translations looks successfully made!

PR #362, opened 2024-03-08
router-fix

### LINKED ISSUE #303: Router Guard to other accounts in someones test
If someone uses your test link in RUXAILAB, he can answer your test without being invited, someone who uses a link like that is like  a 'guest' for the router guard, **fix this to don't allow this.**

skushagra9:
#303

### DIFF
--- a/firestore-debug.log
+++ b/firestore-debug.log
@@ -1,153 +1,46 @@
-Feb 28, 2024 9:51:02 AM com.google.cloud.datastore.emulator.firestore.websocket.WebSocketServer start
+Mar 08, 2024 9:32:02 PM com.google.cloud.datastore.emulator.firestore.websocket.WebSocketServer start
 INFO: Started WebSocket server on ws://127.0.0.1:9150
 API endpoint: http://127.0.0.1:8081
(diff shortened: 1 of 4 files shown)


skushagra9:
I have made the changes in the mods
and the changes in the modified are to help users set it up locally
You can ignore if not needed

skushagra9:
I will add the toast message

src/router/index.js
KarinePistili, line 1: I think the behaviour should be to not be able to answer the test. And show an "unauthorized" warnining.

src/firebase/index.js
KarinePistili, line 1: Please avoid to push unrelated code to the merge request topic. In this case we cannot accept such modification src/firebase/index.js because it would break the overall flow of the app and work only with emulator.

firestore-debug.log
KarinePistili, line 1: Please avoid to push debug logs.

KarinePistili (changes requested):
We would need you to make the changes pointed on the comments above to approve your PR. Also, we would need a visual representation of the warning to users so they understand that they cannot access the test because they were not invited.

PR #425, opened 2024-04-10
Fix Github Actions

sergiobeltranguerrero:
The github Action does not run because the trigger was not well defined.


KarinePistili (approved):

KarinePistili:
Hello @sergiobeltranguerrero !! I wil accept it so we can see it during the next pipelines. I will ping you back if anything goes away from the expected

PR #441, opened 2024-04-17
Issue #435: deleting wrong option in heuristic test

### LINKED ISSUE #435: [🐞 BUG]: deleting wrong option in heuristic test
### Description 📝

When you click on delete the option, it always deletes the last option you have and not the one you want.

### Link 🔗

\src\components\molecules\OptionsTable.vue

### Steps to Reproduce 🔄

- 

### Screenshots 📸

- 

### Expected Behavior 🤔

- 

### Actual Behavior 😱

- 

### Environment 🌍

- 

### Additional Information ℹ️

_No response_

LostSputnik:
### Description

This PR fixes #435 

### What's Causing the Bug

The bug seems to happen exclusively for options with no value set, as discussed [here](https://github.com/ruxailab/RUXAILAB/issues/435#issuecomment-2059213309). To figure out why, first let's take a closer look at the `deleteItem` method:

```
deleteItem(item) {
  this.options.splice(this.options.indexOf(item), 1)
}
```

Now, my understanding is that if there are multiple items with values as null, indexOf might always return the index of the first item with null value, not necessarily the one we want to delete.

### Measures
One way to fix this issue is to assign a unique identifier to each item when it is created, and use this identifier to find the correct item to delete. This ensures that even items with identical visible properties can be correctly distinguished. For this, I have attached a timestamp property to the items. 

### Working Demo

https://github.com/ruxailab/RUXAILAB/assets/34897621/2574367b-413c-4170-8710-fc0a6968128b

### DIFF
--- a/src/components/molecules/OptionsTable.vue
+++ b/src/components/molecules/OptionsTable.vue
@@ -128,6 +128,7 @@ export default {
     },
     updateOptions() {
       if (this.editIndex == -1) {
+        this.option.timestamp = Date.now() // using the current timestamp as a unique identifier
         this.options.push(this.option)
       } else {
         Object.assign(this.options[this.editIndex], this.option)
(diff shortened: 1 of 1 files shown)


KarinePistili:
@FernandaCasali can you review it please?

FernandaCasali (approved):
**Congratulations!** 🎉 Your pull request has been accepted! Thank you for your valuable contribution.

Your changes have been merged successfully and will be included in the develop branch. Your effort in improving our project is greatly appreciated.

Thanks again for your contribution, and we look forward to your continued involvement in our project!

Best regards,
@FernandaCasali

PR #478, opened 2024-05-24
Playwright testing

Zlatanius:
This PR adds adds playwright UI tests, as well as a github action that runs them. This is part of the quality controll subject of which Marc is the professor. Attention should be payed to the change in the Jest configuration. As it stands right now any test contained in a folder called e2e (at any depth) will be ignored by Jest. This change was required because the last configuration was matching any files ending in .spec.js and tried to run playwright tests as unit test and therefore the test GH action would always fail.


sergiobeltranguerrero:
You are using the production url to test the branch. You should use `localhost:port`
Shottr 2024-06-02 10 39 11

Zlatanius:
@sergiobeltranguerrero @marcgc21 requested that the production is tested. I can look into testing development as well sometime later.

sergiobeltranguerrero:
Hi @Zlatanius I'm checking the branch and I'm finding bugs that need to be fixed, at the moment you can't run the tests on a production branch. 

I have also seen that in the tests you log in, but I don't see you register the user previously.....
Shottr 2024-06-02 14 28 20

When I have it sorted out we will approve it and look at doing more tests.

Zlatanius:
Hi @sergiobeltranguerrero, can you explain what do you mean by I can't run tests on the production branch. The branch isn't ready to be tested or do the tests not work?

As for the user, the user used in tests is registered on the system manually.

Zlatanius:
I've added plywright projects that tun tests on the https://ruxailab-dev.web.app/ URL. However there seems to be enough differences that some tests fail on the dev URL. Also I've added video recording to failed tests.

sergiobeltranguerrero:
What I mean is that you are running tests using a URL (https://ruxailab-dev.web.app/) that does not reflect the changes from the current branch.
For example, if you now include a change in a functionality, the GitHub pipeline will fail because that change does not exist in the URL you are using...

Zlatanius:
Ok that makes sense, but production is ok right? My main task is to test the production.

sergiobeltranguerrero:
Yes, the tests are fine. But I can't approve it until I fix the url.

jvJUCA (approved):
Your Pull Request will be approved and your changes will be added to the newest version of develop, and  i will check if everything is working fine!

PR #488, opened 2024-05-28
Ishikawa tools

Zlatanius:
This pull request is part of the Quality Controll class at UdL run by professor Marc. It adds the scripts and the coresponding github action for generation of Ishikawa's diagrams. The workflow is run daily or when there is an issue event and creates a github artefac with the diagrams.


jvJUCA (approved):
Ok, approved!

jvJUCA:
@Zlatanius Hey, I've accepted your pull request and made some fixes with secrets. However, the 'USER' secret represents me, not the repository owner. So, I managed to run the action without errors, but the function that requires 'repo_owner' is still not working.

Zlatanius:
@jvJUCA the problem with the action was that I didn't set the user and repo secrets when made the PR. I see that you have changed it to use an access token. Is it a personal access token? I had changed it to use the GITHUB_TOKEN because it gets generated on every run and isn't time limited.

I think the code should work with the USER secret being set to ruxailab.

PR #489, opened 2024-05-30
Ishikawa tools

Zlatanius:
In this PR I changed back to using the auto generated GITHUB_TOKEN (from testing on a fork this should work). I've also edited the user and project secrets to ruxailab and RUXAILAB repectivly and added a workflow_dispatch trigger to the action so that it can be tested more easily.

### DIFF
--- a/.github/workflows/ishikawa-tools.yml
+++ b/.github/workflows/ishikawa-tools.yml
@@ -1,10 +1,11 @@
 name: Generate Reports
 
-on:
+on: 
   schedule:
     - cron: '0 0 * * *'
   issues:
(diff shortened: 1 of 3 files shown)


sergiobeltranguerrero:
I think the _requirements.txt_ file is missing, because there are libraries like _pandas_, _github_, _matplotlib_ and _requests_ that need to be imported.

Zlatanius:
@sergiobeltranguerrero The dependencies are isntalled in the github action, but can add a requirements file for local testing.

jvJUCA (approved):
I will accept to test on the repository inputs!

PR #514, opened 2024-06-20
small typo fix At testAnswer

saltykheera:


jvJUCA (approved):

jvJUCA (changes requested):
The text has been changed to use internationalization (i18n) for translation. Therefore, if you want to modify the text, you will need to do so in the i18n files.

saltykheera:
> The text has been changed to use internationalization (i18n) for translation. Therefore, if you want to modify the text, you will need to do so in the i18n files.

@jvJUCA this page is in english no matter what language you use . i18n is not emplemented for this page 
image

jvJUCA:
I think you have the old develop

saltykheera:
> I think you have the old develop

i will check

Issue #502, opened 2024-06-13
[✨ FEATURE]: Export Evaluator Answer Table As Csv


saltykheera:
### Feature Description 📝

Add functionality to export data from the Evaluator heuristics table to a CSV file. This feature will allow users to download the table data for external analysis and reporting purposes.

[screen-recorder-thu-jun-13-2024-12-35-10.webm](https://github.com/ruxailab/RUXAILAB/assets/142778969/7c601942-2626-4e70-91ea-2bdfe607888c)

### Motivation 🌟

The current application allows users to view data within the Evaluator heuristics table, but lacks an option to export this data for offline analysis or reporting. By adding CSV export functionality, teams can easily share and analyze data outside the application, facilitating better collaboration and more in-depth analysis

### Expected Behavior 🤔

Users should see an "Export to CSV" button associated with the Evaluator heuristics table. When clicked, the application should generate a CSV file containing the current data in the table and prompt the user to download it. The CSV file should be well-formatted, including appropriate headers for each column in the table.

### Additional Information ℹ️

The CSV export should capture the state of the table as viewed by the user, including any filters or sorting applied.
The implementation should ensure that special characters and commas within the data are properly escaped to maintain CSV integrity.
This feature will enable teams to use external tools like Excel or Google Sheets to create detailed reports and perform advanced data analysis, enhancing the utility of the Evaluator heuristics data.

saltykheera:
@jvJUCA should i create  a PR 
for this feature ?

jvJUCA:
This would be very useful. It would be good to validate whether the data exported as CSV would be useful, and the implementation should be done in a way that ensures quality.

PR #524, opened 2024-07-02
Resolves [✨ FEATURE]: Export Evaluator Answer Table As Csv #502

saltykheera:
Resolves Feature Request #502


jvJUCA (approved):
Accepted, I like this feature and I think it is very useful!

However, I think the style and the way it is presented to the user could be revised.

Thank you for your hard work and dedication to the project!

jvJUCA:
Looking at the code, I see it is based on an older development branch. In your next PR please pull the latest version.

Issue #545, opened 2025-01-19
[✨ FEATURE]: create a user profile page that allows to delete user information and their created tests


marcgc21:
### Feature Description 📝

Until now it is not possible to delete your own user. We need a page accessible through navbar person icon that allows to open a page to see the user information and allow him to delete the user

### Motivation 🌟

It is going to be mandatory to accomplish new policy requirements. 

### Expected Behavior 🤔

I expect to have a new button called "Profile" that will open a new page with the user form. 
One of the expected options will be the "Delete account" that will delete the user from firestore, auth and storage.

<img width="279" alt="Image" src="https://github.com/user-attachments/assets/aaf87cb2-9463-47e7-a9d1-db3725a7cfc8" />

saltykheera:
assign

marcgc21:
Perfect @saltykheera! Let me know if you need something!

saltykheera:
@marcgc21 have a look at my attempt

bobbysharma05:
Can I work on this issue @marcgc21 ?

marcgc21:
We have merged this PR and new issues regarding profile page will be created if you are interested in @bobbysharma05 
I just created one to improve that page.

PR #549, opened 2025-01-19
Fixes  [✨ FEATURE]: create a user profile page that allows to delete …

saltykheera:
Add profile view and delete account functionality; update localization strings with appropriate style and language (i18)
 fixes 
 
image

**For now added a simple Profile Page with ability to delete account on its will with password authentication**
image

steps taken care of 👍 

1. i18n is maintained with appropriate language 
2. icons and styles and adjusted accordingly
3. user can delete the account after the password is entered 

#545


marcgc21:
@saltykheera I started with the review and everthing looks nice! 
Did you check if the user is getting deleted from auth, firestore and storage? I just tried and it looks that it is still not complete. Can you complete that before merging it?

marcgc21:
@saltykheera After reviewing the PR I found that it deletes both firestore and auth. 
However, the user still appeared as logged and it can continue to create testes. We must log out him. After that I beliave that the PR will be completed!

Issue #546, opened 2025-01-19
[✨ FEATURE]: create a notifications view


marcgc21:
### Feature Description 📝

We would like to have a view where the user can see the different notifications that it has received.

### Motivation 🌟

Right now it is not possible to see the complete list of notifications or past notifications. 

### Expected Behavior 🤔

I would like to see a page with the different notifications. We must discuss which fields must appear on that view.

saltykheera:
working on this issue Assign this to me

saltykheera:
https://github.com/user-attachments/assets/5640faf6-1673-4f26-a50d-7383655bd7c6

**Almost done just new sections for all past Notification left to make **
will make the PR on 7:30 AM at Spain time

marcgc21:
I think that as an starting point it will be nice if notifications stops being removed. 
If we think in something like Instagram or Facebook, it is important to mantain the list of notifications received because it might be important at some point to recover and read it again. 

I beliave that it will imply to think a little bit if the database modal do not suit to that functionality. What do you think @saltykheera make it sense for you?

saltykheera:
@marcgc21 
in this pr i have taken care of that 
indox will contain all the notification and will **not** get removed 
whereas 
unread will consists of notification that are not read once read it will get cleard but saved in indox

saltykheera:
https://github.com/user-attachments/assets/416a935d-e41e-46b6-a333-e3e84b166b87

@marcgc21 have a look at this functionality in visual i have implemented in pr #553

marcgc21:
Wow, it looks very promising for a first version! I'll have to test it with my user since I have a lot of messages and it will be a good test of use.

eulerbutcooler:
Can i still take up this issue? @marcgc21

marcgc21:
PR was merged 4 days ago regarding this issue, we are proceeding to close it.

PR #553, opened 2025-01-21
Notification page implemented resolves enhacement request #546 

saltykheera:
#546 
[#546 ](url) 
notification are categeorizes in two section 
A> unread
b> inbox
inbox consist of all the notification user has requested where as 
unread consist of notification which are not read

### DIFF
--- a/src/components/atoms/NotificationButton.vue
+++ b/src/components/atoms/NotificationButton.vue
@@ -1,6 +1,6 @@
 <template>
   <div v-if="user.notifications">
-    <v-menu :position-x="x" :position-y="y" absolute offset-y min-width="200">
+    <v-menu :position-x="x" :position-y="y" absolute offset-y min-width="300">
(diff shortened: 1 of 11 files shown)


marcgc21:
Also, try to update from the develop branch, I think that it solves part of the problems of the CI pipeline!

saltykheera:
> I think that we are moving to a nice update of the system! However, I still have concerns about those notifications dissapearing.
> 
> Can you update the system to avoid cleaning the notifications? Right now I'm not sure if it is a database delete or it is just a filter state on the render visualization. Can you confirm that?

the notification once read are deleted 
and are not saved in database 
hence , the past notifications can't be recovered but after this pr is merged 
the notification will then start to get stored in the database

marcgc21:
Just a video reminder of how it works:

https://github.com/user-attachments/assets/c45c1b7a-b2b9-4257-aee7-9f7770f3bc56

Issue #557, opened 2025-01-23
fix: update requeriments


marcgc21:
After running npm i appears so many legacy dependencies. It might be interesting to start solving there. Maybe it is just a GPT command :)

sahitya-chandra:
@marcgc21 can you assign me this issue

vaishali86c:
@marcgc21 can you assign me this issue

sahitya-chandra:
I am already working on this issue...

gitsofaryan:
we can use dependabot for this.

harith-hacky03:
@marcgc21  I fixed this issue , please check it out https://github.com/ruxailab/RUXAILAB/pull/558 . Let me know if any changes is needed > Happy Contributing : )

marcgc21:
Hi everybody! What we were expecting is to update packages and reduce the number of vulnerabilities. Can someone start working on that?

sahitya-chandra:
@marcgc21 sir i am working on this issue but the problem is that when `ncu -u` or `npm audit fix` it adds some breaking changes.

gitsofaryan:
@marcgc21 im working on reduction.

marcgc21:
@sahitya-chandra yeah I bet that this is the main problem. Handling those crossed dependencies might be hard to solve. Maybe the solution is trying to fix it in small pieces.

hollermay:
@marcgc21 use npm audit fix --force

sahitya-chandra:
@hollermay bro it is not fixing anything but in return it just break the functioning of project.

hollermay:
> @hollermay bro it is not fixing anything but in return it just break the functioning of project.

Then delete the package-lock.json and update the package.json file and the version that solved it for me.

sahitya-chandra:
Done that already. Didn't works.

elenamolero:
Hi, could I contribute?
My name is Elena Molero, and I am excited to apply for the Google Summer of Code with your organization. I am an spanish final-year Computer Engineering student with a strong artistic vision and an extensive portfolio that combines technology and design.

As an active member of the Aula de Software Libre (Free Software Group) at the University of Córdoba (Spain), I have contributed both technically and artistically to various open-source projects and events, including SalmorejoTECH, one of our most recognized initiatives. My experience spans areas such as front-end development, UI/UX design, and creative programming, making me particularly interested in RUXAILAB’s mission to enhance remote usability testing with AI.

[...]

PR #558, opened 2025-01-28
fix : update requirements : #557

harith-hacky03:
**Fixed the issue of displaying legacy dependencies** 
**Before :** 
Screenshot 2025-01-28 at 1 51 36 PM
**After**
Screenshot 2025-01-28 at 1 54 50 PM


marcgc21:
@harith-hacky03 the problem is not appearing the list of deprecated packages what we need is to solve them! Try to work on upgrading some of the packages.

PR #576, opened 2025-02-07
Firebase Database URL Update Issue

gitsofaryan:
# Firebase Database URL Update Issue

## Issue Summary
Previously, Firebase provided the Database URL in the Authentication section. However, Firebase has removed the Database URL from Authentication integration.

## Resolution
Now, the Database URL can be obtained from the Realtime Database section in Firebase Console.

### Steps to Get the Database URL:
1. Go to the [Firebase Console](https://console.firebase.google.com/).
2. Select your Firebase project.
3. Navigate to **Realtime Database** in the left-hand menu.
4. The Database URL is displayed in the top-right corner of the Realtime Database section.

## Fix Implemented
- Updated the documentation to reflect the new location of the Database URL.
- Adjusted the `.env` setup instructions accordingly.

## Additional Notes
Users should now retrieve the Database URL from the Realtime Database section instead of Authentication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,12 @@ UX Remote LAB provides a collaborative environment for creators to share their p
 
 ### Support
 
-- [Report a Bug 🐛](https://github.com/uramakilab/remote-usability-lab/issues/new)
+- [Report a Bug 🦛](https://github.com/uramakilab/remote-usability-lab/issues/new)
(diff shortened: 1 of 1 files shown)


marcgc21:
I was reading it but we're not using Realtime Database. So I bet that it is not precise, I will close it for now and move the discussion to the issue.

PR #583, opened 2025-02-09
Fix Non-.CSV File Upload Issue

amishhaa:
Fixes #582 

This PR adds file type validation to ensure that only .csv files can be uploaded. If a user selects a file with a different extension, an error message is displayed, and the upload process is prevented.

**Fixes**
Prevents users from uploading non-.csv files.
Displays an error message if an invalid file type is selected.
Displays an error message if 'Update' button is pressed while no file is selected.

Screenshot 2025-02-09 212009
Screenshot 2025-02-09 211952

**Changes Made**
Updated changeToJSON() method to check file extension before processing and if the file is uploaded or not.
Implemented error message display for invalid file types.

PR #584, opened 2025-02-10
fix: Fixed the double import of updatePassword function causing the code to break

sahitya-chandra:
In the Last PR #583 @amishhaa added the updatePassword function two times in `src/views/admin/ProfileView.vue` file 
due to this the application is unable to start and it is breaking.

Screenshot from 2025-02-10 18-15-28

Screenshot from 2025-02-10 18-11-14

### DIFF
--- a/src/views/admin/ProfileView.vue
+++ b/src/views/admin/ProfileView.vue
@@ -162,7 +162,6 @@
 import {
   getAuth,
   reauthenticateWithCredential,
-  updatePassword,
   EmailAuthProvider,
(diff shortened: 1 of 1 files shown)


sahitya-chandra:
@marcgc21 @jvJUCA please review this...

amishhaa:
@sahitya-chandra thanks for the fix

jvJUCA (approved):
Thanks for fixing this import issue! I didn't notice it when reviewing @amishhaa's PR.

PR #589, opened 2025-02-12
fix: Updated the deprecated version of actions/checkout and actions/cache

sahitya-chandra:
Updated the deprecated version of actions/checkout and actions/cache to the latest version.
So, now the test.yml workflow will work perfectly.

Screenshot from 2025-02-12 16-19-58


jvJUCA (approved):
thanks for the update

PR #605, opened 2025-02-20
"Save and Leave" Functionality to Ease of usability of software

amishhaa:
Currently there is only one way of saving any kind of changes that are made, that is via. a save button on the bottom right corner,
however that button is hard to find without any context and is not labelled.
When we try to leave a dialogue appears which asks us to stay or leave where all our changes gets discarded, its beneficial to have a feature where we can leave with automatically saving our changes and this PR implements that.
**Before-**
Screenshot 2025-02-20 205002
**After-**
Screenshot 2025-02-20 205015

### DIFF
--- a/src/views/public/TestView.vue
+++ b/src/views/public/TestView.vue
@@ -81,21 +81,29 @@
       <!-- Start Screen -->
       <v-row
         v-if="test && start"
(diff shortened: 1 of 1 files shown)


amishhaa:
@jvJUCA please review

PR #611, opened 2025-02-26
Fixed crucial bug in TestView.vue because of which ModeratedTestView.vue wasn't rendering

amishhaa:
Even after clicking answer test on the dashboard i was not able to attempt the test as the test was not rendered because of incorrect calling of ModeratedTestView.vue in TestView.vue, hence in this PR i fixed the import and the test was rendered correctly.
**Before-**

https://github.com/user-attachments/assets/78325cfc-94dc-4813-ba68-62a74f32d51a

**After-**

https://github.com/user-attachments/assets/c5245c76-aa4f-487a-9e11-920562d85d1a


amishhaa:
@jvJUCA @marcgc21 please review

PR #612, opened 2025-02-26
Update .gitignore to ignore databse-debug.log and ui-debug.log

amishhaa:
while using firebase emulators there is one more debug file that gets created when we create a test which is not included in the .gitignore, this pr adds database-debug.log to .gitignore.
This PR also adds debug-ui.log to .gitignore and also removes ui-debug.log from the codebase.


amishhaa:
@marcgc21 @jvJUCA please review

jvJUCA (approved):

Issue #628, opened 2025-03-05
Enhance Profile Dropdown UI


Va16hav07:
The profile dropdown in Ruxailab has been improved to enhance user experience. The following enhancements have been made:  
- Added a hover effect for better interactivity.  
- Displayed the user’s initial as an icon in the dropdown.  

**Before Enhancement:**  
![Before Enhancement](https://github.com/user-attachments/assets/a0051d05-f5ad-4f79-890a-601a52342e09)  

**After Enhancement:**  
![After Enhancement](https://github.com/user-attachments/assets/55ceaeb9-a536-4425-968c-3a40e5759eb8)

Pranjal6955:
@Va16hav07 Can i work on this issue?

Va16hav07:
@Pranjal6955  This issue is resolved. Please find another.

Pranjal6955:
@Va16hav07 okay

PR #629, opened 2025-03-05
Enhanced Profile Dropdown - Added Hover & Username Initial Icon

Va16hav07:
** Fixes ** : #628  

This PR enhances the profile dropdown in Ruxailab by improving its UI and UX. The changes include:

- Adding a hover effect to make interactions smoother.
- Displaying the user's initial as an icon for a personalized touch.

**Before Enhancement:**  
Before Enhancement  

**After Enhancement:**  
After Enhancement

### DIFF
--- a/src/components/molecules/Toolbar.vue
+++ b/src/components/molecules/Toolbar.vue
@@ -71,7 +71,7 @@
         <v-btn
           dark
           text
-          style="text-transform: none !important; font-family: Roboto, sans-serif; font-size: 17px"
+          class="text-none font-weight-regular text-body-1"
           @click.stop="signOut()"
         >
           {{ $t('buttons.signout') }}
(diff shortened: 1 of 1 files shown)


src/components/molecules/Toolbar.vue
```diff
@@ -452,28 +476,58 @@ export default {
 </script>
 
 <style scoped>
-.console-button {
-  font-size: 13px;
-  font-family: Roboto, sans-serif;
-  text-transform: none !important;
-  padding: 7px !important;
+/* Only keeping styles that can't be handled by Vuetify */
```
amishhaa, line 479: Comments are supposed to be added at places where you want to convey why a specific code was written in a way to help future developers, keeping that in mind, what are you trying to convey with this line?, looks more like a personal note, which could have been written in pr message too.

Va16hav07: The comment says we should use Vuetify instead of inline CSS in future code. This was emphasized to avoid confusion, based on lessons from past PRs. I also found it confusing.

Jamshaa:
Hi @marcgc21  Could you please add me as a reviewer to this PR? Thanks

jvJUCA (approved):
I really liked the changes to the dropdown! The UI looks much better for the user now. Thank you so much for your contribution to the project!

PR #644, opened 2025-03-07
Added 4 language support to finish test page

saltykheera:
finish test
previously it was missing
added for both moderated and unmoderated test


src/views/public/TestView.vue
```diff
@@ -478,7 +468,7 @@
       <UserTestView />
     </div>
     <div v-if="test.testType === 'User' && test.userTestType === 'moderated'">
-      <ModeratedTestView ref="ModeratedTestView" :token="token" />
```
amishhaa, line 481: I opened a pr for fixing this first, so this line was quite unnecessary. @jvJUCA see https://github.com/ruxailab/RUXAILAB/pull/611

marcgc21:
This PRmust be updated with the new json files info

saltykheera:
@marcgc21 resolved

PR #664, opened 2025-03-11
Fix: Prevented  direct mutation of isOpen prop by using computed property

kashi-verma:
In Vue, v-model is syntactic sugar for binding a value to a component and listening for an update event. However, in this code, v-model="isOpen" is directly modifying the isOpen prop, which violates Vue's principle that props should not be mutated directly inside a child component.

### DIFF
--- a/src/components/dialogs/CreateTestNameDialog.vue
+++ b/src/components/dialogs/CreateTestNameDialog.vue
@@ -1,6 +1,7 @@
 <template>
   <div>
-    <v-dialog v-model="isOpen" fullscreen persistent transition="dialog-bottom-transition">
+<v-dialog v-model="dialogOpen" fullscreen persistent transition="dialog-bottom-transition">
+
       <v-card color="#f9f5f0">
         <ButtonBack @click="$emit('close')" />
 
@@ -163,11 +164,20 @@ export default {
     },
   }),
 
-  computed: {
-    user() {
-      return this.$store.getters.user
+ computed: {
+  dialogOpen: {
+    get() {
+      return this.isOpen; // Get the prop value
+    },
+    set(value) {
+      this.$emit('update:isOpen', value); // Emit an event to update the prop in the parent
     },
   },
+},
+// //Why This Works?
+// Getter (get): Returns the current value of isOpen (the prop).
+// Setter (set): Emits an event (update:isOpen) to inform the parent that isOpen has changed.
+// Parent Component: Must listen for update:isOpen and update isOpen accordingly.
 
   methods: {
     validate() {


src/components/dialogs/CreateTestNameDialog.vue
```diff
@@ -163,11 +164,20 @@ export default {
...
 
-  computed: {
-    user() {
-      return this.$store.getters.user
+ computed: {
+  dialogOpen: {
+    get() {
+      return this.isOpen; // Get the prop value
+    },
+    set(value) {
```
amishhaa, line 172: I think the intended behavior was READ ONLY, so set method was not required, also as it is read only there is no Mutating of a prop. Can you provide more details on what the error was? and where the previous code was lagging?

kashi-verma: You're correct that v-model="isOpen" directly modifying a prop violates Vue’s principle that props should not be mutated inside a child component. In Vue 3, v-model is shorthand for:

:someProp="someValue" @update:someProp="someValue = $event"
Since isOpen is a prop, directly using v-model on it will try to mutate it, which is not allowed.

Solution
Instead of using v-model="isOpen" directly, use a computed property with a getter and an event emitter for updates:

Fixed Code (Child Component - CreateTestNameDialog.vue)

<script setup>
import { computed, defineEmits, defineProps } from "vue";

const props = defineProps({
  isOpen: Boolean
});

const emit = defineEmits(["update:isOpen"]);

const openDialog = computed({
  get() {
    return props.isOpen; // Get the prop value
  },
  set(value) {
    emit("update:isOpen", value); // Emit event instead of mutating prop
  }
});
</script>

<template>
  <Dialog v-model="openDialog">
    
  </Dialog>
</template>
Usage in Parent Component
vue
Copy
Edit
<CreateTestNameDialog v-model:isOpen="dialogVisible" />
This way, the parent controls dialogVisible, and the child correctly emits an event when it needs to update the value.

Why This Fix Works?
Props are immutable in child components → This ensures we don’t mutate props directly.
Vue’s v-model convention is preserved → Now the parent properly handles state changes.
Computed setter emits an event → Instead of modifying isOpen, it notifies the parent to update it.

amishhaa: > You're correct that v-model="isOpen" directly modifying a prop violates Vue’s principle that props should not be mutated inside a child component. In Vue 3, v-model is shorthand for:

We use vue2 not vue3.

kashi-verma: The issue here is that isOpen is a prop, and in Vue, props are meant to be read-only inside child components. Mutating isOpen directly inside CreateTestNameDialog.vue violates Vue’s best practices.

If the previous version worked "flawlessly," it might have been because Vue 2 didn’t enforce the prop immutability strictly, or it was working but still technically incorrect.

amishhaa: > The issue here is that isOpen is a prop, and in Vue, props are meant to be read-only inside child components. Mutating isOpen directly inside CreateTestNameDialog.vue violates Vue’s best practices.

i understand that, but as i said earlier there is no mutation as the props are read only so whats the problem? also you are adding a fix for a problem which you are introducing yourself, if there is not set component there is no mutation.

kashi-verma: I understand your point that there is no direct mutation since isOpen is a read-only prop and there’s no set method. However, the issue arises due to how v-model works in Vue 2.


v-model="isOpen" in a child component assumes two-way binding

In Vue 2, v-model is shorthand for:
:value="isOpen" @input="isOpen = $event"
Since isOpen is a prop, Vue expects the parent to handle the @input event and update the value accordingly.
If the parent does not handle the update, Vue might internally treat this as an attempt to modify the prop directly, leading to warnings or unexpected behavior.
If there is no @input event emitted, v-model won’t work correctly

Even though the prop is read-only, Vue might still raise concerns if the parent isn’t receiving updates from the child.

amishhaa: > there’s no set method. 

Currently there's none, you are introducing one in your code, what is the reason for that? 

>If the parent does not handle the update, Vue might internally treat this as an attempt to modify the prop directly, leading to warnings or unexpected behavior.
If there is no @input event emitted, v-model won’t work correctly

Did you try to reproduce this error?

kashi-verma: You're right—there's no set method currently, and we're introducing it not to mutate the prop, but to properly handle Vue 2’s v-model behavior.

In Vue 2, v-model="isOpen" is shorthand for:
:value="isOpen" @input="isOpen = $event"
Since isOpen is a prop, Vue expects an @input event for updates. Without it, Vue might internally treat this as a mutation, leading to warnings.

The computed setter ensures we emit an event instead of mutating the prop, keeping the parent in control:
computed: {
  openDialog: {
    get() { return this.isOpen; },
    set(value) { this.$emit("input", value); }
  }
}
Did you try reproducing the error? If Vue didn’t warn earlier, it might not have been strictly enforcing prop immutability, but this is the correct approach for proper reactivity.

marcgc21: So what is the final veredict? We need to add it or not?

marcgc21: @kashi-verma, are you agree with that? Can I close it and let it go for those that will work on the implementation of vue3?

kashi-verma: @marcgc21 I understand the concerns raised regarding this change. My intention was to align with best practices by ensuring that props remain immutable inside child components and that updates are handled correctly via event emissions.

While Vue 2 does not enforce prop immutability as strictly as Vue 3, directly using v-model="isOpen" still assumes an internal mutation unless we explicitly emit an event. However, if the current implementation is functioning correctly without any warnings or unexpected behavior, then the change may not be necessary at this moment.

That being said, adopting the event-based approach (@input emission) ensures a seamless transition when migrating to Vue 3 in the future. If the team decides to stay with Vue 2 for now and the existing implementation does not introduce any issues, I am okay with closing this for now.

marcgc21:
We are going to close PR due to inactivity :(

Issue #671, opened 2025-03-12
[✨ FEATURE]: Add validation message on signUp form


adizh:
### Feature Description 📝

Add validation message on signup form button. The validation appears only after users start typing, on form submission it checks, but without messages.

<img width="614" alt="Image" src="https://github.com/user-attachments/assets/23b8ec69-b683-4c21-8432-9d7c13d61cbe" />

### Motivation 🌟

Better user experience

### Expected Behavior 🤔

Validation messages should appear on form submission also.

amaan-malik1:
Show validation messages dynamically as users type (after they start typing).
Ensure validation runs on form submission and prevents invalid data from being submitted.

i would love to contribute to this...

adizh:
@amaan-malik1 this issue is resolved, I am waiting for PR to be approved

PR #672, opened 2025-03-12
Validation for signup form

adizh:
Added validation check to signup form on submit.

The messages will appear on form submit also.

Screenshot 2025-03-12 at 11 16 37

#671


marcgc21:
Same as the other PR, so many files when only one file must be changed.

Issue #666, opened 2025-03-11
[🐞 BUG]: On error /signup page directly redirects to home page,


adizh:
### Description 📝

On some error /signup page directly redirects to home page. If email is already in use, the Toast message appears after the user is redirected to home page. It not comfortable, because user needs to again to signup
page

### Link 🔗

/signup

### Steps to Reproduce 🔄

1. Go to signup
2. Type email that is already in use
3. The website will navigate you to home page

### Screenshots 📸

<img width="1006" alt="Image" src="https://github.com/user-attachments/assets/fda501b0-b2ff-4c8a-a2fc-f3581279042f" />

### Expected Behavior 🤔

User should not be redirected to home page, if any error occurs in /signup page. Only in case of successful auth they should be navigated to home page

### Actual Behavior 😱

Users are being navigated to home page on any error occurring in /signup page

### Environment 🌍

-Vue components: SignUpView.vue

aniket866:
@adizh  Want to solve please assign me 👍

adizh:
@aniket866  also working on this

suleman1412:
Hi @adizh , working on this

suleman1412:
hi @adizh please take a look at the PR #667 and let me know if any changes need to be made

marcgc21:
Solved

PR #673, opened 2025-03-12
Fix redirect to home page on signup error

adizh:
Fixed a navigation in signup page, on error it will stay in the same page, only in case of successful signup it will navigate the user to home page

Screenshot 2025-03-12 at 12 02 30

#666

### DIFF
--- a/src/firebase/index.js
+++ b/src/firebase/index.js
@@ -13,8 +13,8 @@ const firebaseConfig = {
   projectId: process.env.VUE_APP_FIREBASE_PROJECT_ID || '',
   messagingSenderId: process.env.VUE_APP_FIREBASE_MESSAGING_SENDER_ID,
   appId: process.env.VUE_APP_FIREBASE_APP_ID,
-
 }
+
 
 
 const firebaseApp = initializeApp(firebaseConfig)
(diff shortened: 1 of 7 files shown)


src/firebase/index.js
```diff
@@ -24,11 +26,11 @@ const storage = getStorage(firebaseApp, `gs://${firebaseConfig.storageBucket}`)
 
 // Emulators if running locally
 
-// if (process.env.NODE_ENV === 'development') {
-//   connectFirestoreEmulator(db, 'localhost', 8081)
-//   connectAuthEmulator(auth, 'http://localhost:9099')
-//   connectFunctionsEmulator(fbFunctions, 'localhost', 5001)
-//   connectStorageEmulator(storage, '127.0.0.1', 9199)
-// }
+if (process.env.NODE_ENV === 'development') {
```
amishhaa, line 29: you are not supposed to push your code with these lines uncommented make a commit adding this file name to .gitignore and then push this.

amishhaa:
I went through all of your pull requests and they are really messy as you have made all changes in one branch, i suggest you to make separate branches for separate changes you want to make in a pr that way your changes in one pr wont reflect in the next one.

adizh:
@amishhaa  they were pushed  in separate branches

amishhaa:
> @amishhaa they were pushed in separate branches

That's good, but all of your PR has the same bugs in the same files. Those unnecessary lines you were changing, if you had made seperate branches then you have to make a commit fixing that in every branch.

src/firebase/index.js
adizh: @jvJUCA can you please clarify on this, does this file should be in gitignore?

marcgc21:
I recently solve it, it was just a matter of two lines of code.

PR #675, opened 2025-03-12
Fix: Improved Error Handling in update Method

### LINKED ISSUE #674: Improve Error Handling in update Method
The update method in the Controller class currently has a try-catch block that only logs the error when an exception occurs. However, just logging the error is not enough, as it does not notify the caller about the failure. This can lead to silent failures where the function call does not execute properly but the user is unaware of it.

Current Implementation (Problematic Code):

async update(col, docId, payload) {
  try {
    const ref = doc(db, `${col}/${docId}`)
    return updateDoc(ref, payload)
  } catch (e) {
    console.log(e) // Only logs the error, does not propagate it
  }
}

Why is this an Issue?
1. Lack of proper error handling: The error is only logged, so the calling function will not know whether the update was successful or not.
2. No feedback to the user: The application will continue running without informing the user about the failure.
3. String interpolation is unnecessary in doc(db, ${col}/${docId}) – Firestore's doc() function already accepts multiple arguments.

**Suggested Fix:**
The catch block should throw an error or return an error response instead of just logging it.

Image

kashi-verma:
Fix Description:
- Improved error handling in `update` method.
- Instead of just logging the error, the function now throws an error.
- Removed unnecessary template literals when calling `doc()`.

Fixes #674

### DIFF
--- a/src/controllers/BaseController.js
+++ b/src/controllers/BaseController.js
@@ -117,14 +117,16 @@ export default class Controller {
     return res
   }
 
-  async update(col, docId, payload) {
-    try {
-      const ref = doc(db, `${col}/${docId}`)
-      return updateDoc(ref, payload)
-    } catch(e) {
-      console.log(e)
-    }
+ async update(col, docId, payload) {
+  try {
+    const ref = doc(db, col, docId) // Removed unnecessary template literal
+    return await updateDoc(ref, payload)
+  } catch (e) {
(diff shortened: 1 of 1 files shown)


sahitya-chandra (changes requested):
There is no need to throw an error when you are also logging it.
This just increases the redundancy.

marcgc21:
We are closing it due to no activity!

PR #702, opened 2025-03-15
Fix: prevent navigation error when prevRoute is undefined  

### LINKED ISSUE #701: [🐞 BUG]: PrevRoute Handling Edge Cases
### Description 📝

The prevRoute variable stores the previous route and is used in the sendHome() method to navigate back to the last visited page. However, if prevRoute is null (which is handled correctly) or prevRoute.path is undefined, attempting to access it (this.prevRoute.path) can result in an error.

### Link 🔗

https://github.com/ruxailab/RUXAILAB/blob/develop/src/views/public/PageNotFoundView.vue

### Steps to Reproduce 🔄

1. Navigate directly to the "Page Not Found" component (e.g., by entering the URL manually).
2. Click the "Go Back" button.
3. Since there is no previous route (prevRoute.path is undefined), the application may throw an error.

### Screenshots 📸

Image

### Expected Behavior 🤔

1. When the user navigates to a non-existent page, they see the "Page Not Found" screen.
2. Clicking the "Go Back" button should take them to the previously visited page.
3. If there is no previous route (e.g., when the user lands directly on the 404 page via a URL), clicking "Go Back" 
    should redirect them to the home page ('/') without errors.

### Actual Behavior 😱

1. If the user lands directly on the "Page Not Found" page and clicks the "Go Back" button, prevRoute may be null 
    or undefined.
2. Since prevRoute.path is accessed without checking if it exists, it can lead to a runtime error (Cannot read property   
    'path' of undefined).
3. This may result in an unhandled exception or unexpected behavior instead of smoothly redirecting the user.

### Environment 🌍

- Vue component: `src/components/PageNotFound.vue`
- Git branch: `main`
- OS: Windows 11
- Browser: Chrome
- Version: Vue 2.x / Vuetify 2.x (or the specific versions you are using)

### Additional Information ℹ️

This issue could affect the user experience by causing navigation failures when accessing the "Page Not Found" page directly.

kashi-verma:
- Added a safe check to ensure prevRoute exists and has a valid path before accessing it.  
- Prevents runtime errors when clicking "Go Back" on the 404 page after direct navigation.  

Fixes #701

### DIFF
--- a/src/views/public/PageNotFoundView.vue
+++ b/src/views/public/PageNotFoundView.vue
@@ -27,12 +27,22 @@ export default {
   data: () => ({
     prevRoute: null,
   }),
-  methods: {
-    sendHome() {
-      if (this.prevRoute !== null) this.$router.push(this.prevRoute.path).catch(() => {})
-      else this.$router.push('/').catch(() => {})
-    },
+ methods: {
+  sendHome() {
+    // Check if prevRoute exists and has a valid path before attempting navigation
+    if (this.prevRoute && this.prevRoute.path) {
+      this.$router.push(this.prevRoute.path).catch(() => {
+        // Catch any navigation errors to prevent unexpected crashes
+      });
+    } else {
+      // If no previous route exists, redirect the user to the home page
+      this.$router.push('/').catch(() => {
+        // Catch any navigation errors to prevent unexpected crashes
+      });
+    }
   },
+},
+
   beforeRouteEnter(to, from, next) {
     next((vm) => {
       vm.prevRoute = from


src/views/public/PageNotFoundView.vue
```diff
@@ -27,12 +27,22 @@ export default {
   data: () => ({
     prevRoute: null,
   }),
-  methods: {
-    sendHome() {
-      if (this.prevRoute !== null) this.$router.push(this.prevRoute.path).catch(() => {})
```
amishhaa, line 32: this.prevRoute object is only created when a path exists so if this.prevRoute is not null this.prevRoute.path is also not null. This code was correct before too, did you try to reproduce the error, did any instance come up when redirecting failed?

marcgc21:
Not working

PR #705, opened 2025-03-15
Added chips for access level implemented Color theory

### LINKED ISSUE #571: feat(admin): improvement of user lists
As a super administrator we need a view where to find the different users and their current status.
Some table like this must be awesome. 
We imagine that columns will be:
- User
- Role
- Num. tests created
- Status
- Actions

Filters might be just search by name to make it easier.

Regarding actions we need two: delete user and change their current status to inactive

Image

saltykheera:
Fixes #571 
implemented Right color theory chips 
image


marcgc21:
Like it!

Issue #727, opened 2025-03-19
[🐞 BUG]:  Inconsistent show State Behavior


kashi-verma:
### Description 📝

Currently, the show variable determines whether the comment input box is visible. It toggles when the user clicks on the comment button. However, when the heurisIndex prop changes (indicating that the user has switched to a different heuristic), the watch property forces show to reset to false:

watch: {
  heurisIndex() {
    this.show = false //close comment when changing heuristic
  },
},

The problem here is that:

* The user's existing comment remains stored in answerHeu.heuristicComment, but the text area is hidden when 
   switching heuristics.
* This creates a poor user experience, as the user may believe their comment has been lost.
   The UI behaves unexpectedly because the comment exists but is no longer visible.

🎯 Impact Analysis
User Confusion:
1. If a user enters a comment but then selects a different heuristic, they may think their previous comment was 
   erased, even though it still exists in answerHeu.heuristicComment.
2. Poor UI Experience:
  The sudden disappearance of the comment text box makes the UI feel unresponsive or broken.

### Link 🔗

https://github.com/ruxailab/RUXAILAB/blob/develop/src/components/atoms/AddCommentBtn.vue

### Steps to Reproduce 🔄

1. Go to [page where the issue occurs].
2. Click on the comment button to add a comment.
3. Enter a comment and then change the heuristic selection.
4. Observe that the comment text box disappears, even though the comment is still stored in the data model.

### Screenshots 📸

![Image](https://github.com/user-attachments/assets/cb1641ab-520b-41a7-bb8a-1cfb3d74ca84)

### Expected Behavior 🤔

If a comment exists, the comment text box should remain visible when switching heuristics.
The UI should clearly indicate that the comment is saved and accessible.

### Actual Behavior 😱

The comment section disappears when switching heuristics.
The user might believe their comment is lost, even though it is still stored.

### Environment 🌍

Vue component: src/components/HeuristicComment.vue (or the relevant file)
Git branch: main (or the branch where the issue occurs)
OS: Windows 10 / macOS / Linux (specify your OS)
Browser: Chrome / Firefox / Safari (mention the browser used)
Version: (Specify the project or application version)

chanchal430:
Hi 👋 @kashi-verma , 

I have tested the behavior after the latest updates. The issue regarding the comment box disappearing unexpectedly when switching heuristics seems to be fixed now. 

✅ The comment box stays visible when there is a saved comment or uploaded image.
[...]

PR #728, opened 2025-03-19
Update AddCommentBtn.vue

kashi-verma:
🔹 Step 1: Remove show from data() and watch block
Instead of relying on data() for show, we will replace it with a computed property.

🔹 Step 2: Define a computed property shouldShowComment
Modify the script to include:

computed: {
  shouldShowComment() {
    return this.answerHeu.heuristicComment?.trim().length > 0 || this.show;
  },
},

This ensures that:

If answerHeu.heuristicComment contains a non-empty string, the comment box remains visible.
Otherwise, it follows the show state.

🔹 Step 3: Use shouldShowComment in the Template
Replace all instances of v-if="show" with v-if="shouldShowComment":

<v-textarea
  v-if="shouldShowComment"
  v-model="answerHeu.heuristicComment"
  outlined
  dense
  auto-grow
  clearable
  clear-icon="mdi-close"
  :label="$t('common.comment')"
  @change="updateComment"
/>

<ImageImport
  v-if="shouldShowComment"
  :heuristic-id="test.testStructure[heurisIndex]"
  :question-id="answerHeu.heuristicId"
  :test-id="this.$store.getters.test.id"
  @imageUploaded="handleImageUploaded"
/>

🔹 Step 4: Adjust the @click Event to Toggle show
Modify the button's @click event to ensure it updates show correctly:

<v-btn icon @click="show = !show">
  <v-icon>mdi-comment-plus-outline</v-icon>
</v-btn>
#727

### DIFF
--- a/src/components/atoms/AddCommentBtn.vue
+++ b/src/components/atoms/AddCommentBtn.vue
@@ -5,9 +5,9 @@
         <slot name="answer" />
       </v-col>
       <v-col cols="1" class="mb-6 py-0">
-        <v-tooltip v-if="!show" bottom>
+        <v-tooltip v-if="!shouldShowComment" bottom>
           <template v-slot:activator="{ on, attrs }">
-            <v-btn icon v-bind="attrs" @click="show = !show" v-on="on">
+            <v-btn icon v-bind="attrs" @click="toggleComment" v-on="on">
               <v-icon :color="answerHeu.heuristicComment ? '#F9A826' : ''">
                 mdi-comment-plus-outline
               </v-icon>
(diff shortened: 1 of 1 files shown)


Jordipifarre:
Hello @marcgc21 please add me as reviewer here.

amishhaa:
I think the previous behaviour was the same, you are just adding new variable names, can you add screenshots of previous and new behaviour so validating the PR becomes easier? Because right now it doesn't look like any change has occured to me.

Jordipifarre:
Thanks for the PR. I think using the computed property shouldShowComment instead of relying directly on show state is a solid improvement — it centralizes logic and improves code readability.

That said, I noticed the watch on heurisIndex that previously reset show to false when switching heuristics has been removed. I'm assuming this is intentional, but just to confirm:
Is the idea now that if the comment area is open, it stays open even when switching to a different heuristic?

Just pointing it out because it changes the previous UX behavior — and it might cause confusion if the comment box remains visible but no longer matches the currently selected heuristic.

Other than that, I’m aligned with the proposed refactor. I also second @amishhaa’s suggestion to include before/after screenshots to better validate the impact. Thanks!

PR #738, opened 2025-03-22
Optimize watcher for snackMessage to avoid redundant calls

kashi-verma:
1. Updated watch.snackMessage() to use the newVal parameter 
    instead of calling this.snackMessage again.
2. This improves efficiency by preventing duplicate computed 
    property calls.
3. No changes in functionality—just a performance improvement.
    No breaking changes; the component behavior remains the 
    same.

Why This Change Matters?
🔹 Improves performance by avoiding unnecessary reactivity 
      updates.
🔹 Ensures Vue evaluates snackMessage only once per update.
🔹 Keeps the code cleaner and more readable.

### DIFF
--- a/src/components/atoms/Snackbar.vue
+++ b/src/components/atoms/Snackbar.vue
@@ -27,12 +27,12 @@ export default {
       return this.$store.getters.snackColor
     },
   },
-  watch: {
-    snackMessage() {
-      if (this.snackMessage) this.snackbar = true
+watch: {
+    snackMessage(newVal) {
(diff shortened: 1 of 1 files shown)


amishhaa:
@kashi-verma you're complicating the codebase unnecessarily by adding another variable, if you intend to add something to improve the performance make sure you then run it on some benchmark to prove that performance has increased instead of trying out stuff blindly. @marcgc21 correct me if I am wrong thank you.

PR #746, opened 2025-03-24
Validation added for signup page

adizh:
Form validation added for sign up page on submit


7MCristy:
hi @marcgc21 can you add me as a reviewer comment , please?

7MCristy:
hi @adizh, there's some conflicts to merge, can you solve it?

adizh:
@7MCristy conflicts resolved

adizh:
@7MCristy @marcgc21  there are build errors:  if [ -z "${{ secrets.VUE_APP_FIREBASE_DB_URL }}" ]; then
            echo "❌ Error: Missing secret VUE_APP_FIREBASE_DB_URL"
            missing=true
          fi
          What does this mean? I have these keys inside .env

7MCristy:
Hi @adizh ! The problem it is that it's not in the correct place the .env, but you won't need to deal with that . 
However, I wanted to ask if you would consider the explicit error catching in the first code snippet as a better practice, especially in terms of user clarity and providing user-friendly error messages are nice to give information.

7MCristy (commented):
Hi @adizh ! The problem it is that it's not in the correct place the .env, but you won't need to deal with that .
However, I wanted to ask if you would consider the explicit error catching in the first code snippet as a better practice, especially in terms of user clarity and providing user-friendly error messages are nice to give information.

adizh:
Hello @7MCristy, the user friendly messages were already provided(like email validation, the field is required etc), it just on submit there was no validation, validation appeared only after users start typing.  If you mean this

7MCristy:
@adizh well, I wanned to make reference to the try catch and the message that it's sent to the user in case of error so the code has the validation and the descriving error message, in the modified code, or there's a bigger reason why you delete it?

adizh:
@7MCristy I did not delete anything from the code, just added validation. These are the messages that will appear on submit 
Screenshot 2025-03-29 at 11 03 37

adizh:
@marcgc21 I did not delete that on purpose. At that time when I switched to develop branch this is how it was, that is, perhaps it happened when I was resolving conflicts, even then I did not remember deleting that try/catch block
Screenshot 2025-03-29 at 14 38 31

PR #797, opened 2025-04-14
Added Enhacements for manual accessibility testing

saltykheera:
submissions now include this missing parameters .

1. A field is missing to post the screenshot image of the evidence of the problem.
2. It is necessary to add a field per notes or description of the appraiser.
3. Lack of information as to the details of the report in PDF
4. It is necessary to be able to visualize the patterns in two ways (or one by one) or all of the copy.
5. I also see the information, not with the folded tabs, but rather the results without having to click.


saltykheera:
@marcgc21 i have added the requested enhacements

PR #811, opened 2025-04-17
Update Documentation.vue

Sarthak-Developer-Coder:
Key Improvements Made:
Data Pre-processing:

Added a processedPrinciples computed property that flattens the data structure once

This eliminates repeated Object.values() calls in the template

Better Keying:

Used proper unique keys for all v-for loops

Added id fields where missing in the processed data

Performance Optimizations:

Removed all Object.values()[0] calls from the template

Computed properties are cached by Vue's reactivity system

Data is processed only once when dependencies change

Code Organization:

Renamed wcagData to rawWcagData to be more explicit

Kept the original styling and layout intact

Maintained all existing functionality

### DIFF
--- a/src/views/public/Documentation.vue
+++ b/src/views/public/Documentation.vue
@@ -13,18 +13,18 @@
             multiple
           >
             <v-expansion-panel
-              v-for="(principle, index) in wcagData"
-              :key="index"
+              v-for="(principle, index) in processedPrinciples"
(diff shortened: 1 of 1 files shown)


sahitya-chandra:
@Sarthak-Developer-Coder, why have you raised 2 PRs for the changes in same file ??

PR #812, opened 2025-04-20
🔧 Change: Replace template v-for with v-list-item loop directly

kashi-verma:
✅ Summary of Changes:
    1. Replaced the <template v-for> wrapper with a direct v-for loop on the <v-list-item> and <v-divider> elements.
    2. Ensured each notification and its divider are grouped logically and rendered cleanly with proper key binding.
    3. Added title and aria-label to action icons for better accessibility (a11y).
    4. Maintained original logic and layout while improving overall code readability.

❌ Why it's suboptimal:
     1. Wrapping v-for in a <template> is usually used when you don’t want to render an actual HTML element, but here you're 
        always rendering a real element (<v-list-item> and <v-divider>).
     2. It can confuse the rendering logic, and affects readability for larger teams or contributors.
     3. It doesn’t clearly group each notification and its divider together

✅ Recommended Change:
Use v-for directly on a <v-fragment> (or a <div> if using Vue 2.x with Vuetify 2.x) to group each list item and its divider clearly:

✅ Updated Code:

<template>
  <v-list two-line>
    <template v-for="(notification, index) in notifications" :key="notification.id">
      <v-list-item @click="$emit('go-to-redirect', notification)">
        
      </v-list-item>

      
      <v-divider v-if="index < notifications.length - 1" />
    </template>
  </v-list>
</template>

✅ Benefits:

1. Cleaner, more intuitive loop structure
2. Avoids unnecessary <template> nesting
3. Easier for other devs to read and maintain
4. Ensures correct key usage at the right element level

### DIFF
--- a/src/components/molecules/NotificationList.vue
+++ b/src/components/molecules/NotificationList.vue
@@ -1,14 +1,9 @@
 <template>
   <v-list two-line>
-    <template v-for="(notification, index) in notifications">
-      <v-list-item
-        :key="notification.id"
-        @click="$emit('go-to-redirect', notification)"
-      >
+    <template v-for="(notification, index) in notifications" :key="notification.id">
+      <v-list-item @click="$emit('go-to-redirect', notification)">
         <v-list-item-content>
-          <v-list-item-title
-            :class="{ 'font-weight-bold': !notification.read }"
-          >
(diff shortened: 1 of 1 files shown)


marcgc21:
This PR do not solves that still have harcoded strings that muat be placed on i10n. As we are working on the refactor of vue 3 we prefer to close that.

Issue #802, opened 2025-04-15
[🐞 BUG]:  Performance & Re-rendering Problem Due to Use of Object.values(principle)[0] in Template


kashi-verma:
### Description 📝

Description: 
In the current template, we repeatedly use Object.values(principle)[0] within the v-for. This is inefficient and leads to unnecessary re-renders and CPU usage, especially for large datasets. We should extract this value once and use a cleaner reference (principle) for better performance and readability.

* Vue has to re-evaluate Object.values(principle)[0] every time the component re-renders.
* If your dataset is large, this causes performance degradation and unnecessary computation.
* Also it makes the template harder to read and debug.

### Link 🔗

https://github.com/ruxailab/RUXAILAB/blob/develop/src/views/public/Documentation.vue

### Steps to Reproduce 🔄

Open the component WcagDocumentation.vue.
Load the UI containing the nested <v-expansion-panels> rendering WCAG principles.
Inspect or monitor performance via browser DevTools.
Notice repeated re-computations when interacting with the panel — especially if the dataset is large.

### Screenshots 📸

Not required to reproduce, as the issue is performance-related and visible in the code and via profiler.

### Expected Behavior 🤔

The component should render efficiently, accessing data directly via clear object properties (e.g., principle.title, principle.description) without requiring repeated calls to Object.values() in the template. This would improve performance and maintainability.

### Actual Behavior 😱

Currently, the template repeatedly uses expressions like Object.values(principle)[0] across multiple bindings:

{{ Object.values(principle)[0].title }}

This introduces:
1. Unnecessary re-evaluation during reactivity updates.
2. Poor performance for large datasets.
3. Reduced readability and maintainability.

### Environment 🌍

Vue component: src/components/WcagDocumentation.vue
Git branch: main
OS: Any (issue is frontend logic-related)
Browser: Any (observed on Chrome)
Vuetify: (e.g., 2.x or 3.x depending on your setup)
Vue version: (e.g., 2.x or 3.x)

kashi-verma:
SUre @chanchal430, this is available and you can absolutely work on this issue_

chanchal430:
Thank you! will get back once it is completed @kashi-verma

PR #813, opened 2025-04-22
Optimize WCAG documentation rendering for performance and readability

chanchal430:
### **Summary**
This PR improves the performance and readability of the **`Documentation.vue`** component, which displays **WCAG accessibility guidelines**.

### **fixes: #802** 
Previously, the template frequently accessed deeply nested data using **Object.values(principle)[0]** inside the **v-for loop**. This resulted in:
- Redundant computations on each re-render
- Increased CPU usage and slower rendering for large datasets
- Less readable and harder-to-maintain template code

### **✅ What was done**
- Refactored **v-for="(principle, index) in wcagData"** to extract the actual **principle data** before using it in the template
- Replaced all **Object.values(principle)[0**] calls with a direct reference.
- Improved semantic clarity and reduced rendering cost

#### **🔬 Before optimization the performance Snapshot:**

before

#### **Screenshots of DevTools performance snapshots**

performance-before

#### **⚡ After optimization the performance Snapshot:**

after

#### **Screenshots of DevTools performance snapshots**

performance-after

### **🚀 Why this matters**
This optimization becomes especially important as the dataset scales. For example, if more WCAG principles, guidelines, or rules are added:
- Vue will avoid repeating the expensive Object.values() computation per item
- Component re-renders will remain smooth even with significantly more content
- Overall performance will scale better with growing data size

### **🔎 How to test**
- Run the app locally
- Navigate to /wacg
- Open DevTools → Performance tab
- Interact with expansion panels, observe render speed and memory usage

### **🎯 Outcome**
- Faster initial render (LCP reduced by ~250ms)
- Significantly reduced scripting and rendering time
- Improved maintainability and readability of the template

This change will benefit users on slower devices and contribute to a smoother user experience.

### DIFF
--- a/src/views/public/Documentation.vue
+++ b/src/views/public/Documentation.vue
@@ -2,18 +2,19 @@
   <v-container class="container-max-width">
     <v-row justify="center">
       <v-col cols="12" class="text-center mb-6">
-        <h1 class="display-1 font-weight-bold primary--text">WCAG Accessibility Guidelines</h1>
-        <div class="subtitle-1 grey--text text--darken-1">Web Content Accessibility Guidelines Documentation</div>
+        <h1 class="display-1 font-weight-bold primary--text">
+          WCAG Accessibility Guidelines
+        </h1>
+        <div class="subtitle-1 grey--text text--darken-1">
+          Web Content Accessibility Guidelines Documentation
+        </div>
       </v-col>
-      
+
       <v-col cols="12">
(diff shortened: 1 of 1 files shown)


chanchal430:
Hi @marcgc21 👋

This PR is ready for review — it optimizes the WCAG documentation rendering for better performance and scalability. I've added before/after performance snapshots for reference. Would appreciate your feedback whenever you get the chance!

Thanks 🙌

PR #815, opened 2025-04-24
Update DisconnectedCard.vue

kashi-verma:
Replaced overlapping if conditions in updateDialogMaxWidth() with a clean if-else if-else chain to ensure mutually exclusive execution. This now correctly applies 80vw for mobile, 60vw for tablets, and 40vw for desktops, improving the component's responsiveness and expected behavior.
The method updateDialogMaxWidth() is supposed to adjust the dialog's width based on the screen size:

if (window.innerWidth <= 600) {
  this.dialogMaxWidth = '80vw'
}
if (window.innerWidth <= 900) {
  this.dialogMaxWidth = '60vw'
} else {
  this.dialogMaxWidth = '40vw'
}

❌ Why this is incorrect:
These are independent if statements, so both can execute.

For window.innerWidth = 500, both conditions <= 600 and <= 900 are true.

So, dialogMaxWidth is first set to 80vw, then overwritten to 60vw.

The intended value 80vw for small devices is never used.

✅ Solution: Use if ... else if ... else
This ensures only one block runs depending on screen width, fixing the logic bug:

updateDialogMaxWidth() {
  if (window.innerWidth <= 600) {
    this.dialogMaxWidth = '80vw' // for phones
  } else if (window.innerWidth <= 900) {
    this.dialogMaxWidth = '60vw' // for tablets
  } else {
    this.dialogMaxWidth = '40vw' // for desktops
  }
}

### DIFF
--- a/src/components/atoms/DisconnectedCard.vue
+++ b/src/components/atoms/DisconnectedCard.vue
@@ -10,10 +10,10 @@
       <v-container fluid>
         <v-row no-gutters>
           <v-col class="text-center cardSubtitle">
-            <span
-              >Infelizmente ocorreu uma desconexão. Você precisará realizar a
-              conexão novamente!</span
-            >
+            <span>
+              Infelizmente ocorreu uma desconexão. Você precisará realizar a
+              conexão novamente!
+            </span>
           </v-col>
           <v-col class="text-center">
             <v-img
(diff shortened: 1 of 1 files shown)


marcgc21:
This PR should place the strings to other languages, i will close it

PR #838, opened 2025-05-01
Update TipButton.vue

kashi-verma:
🚨 Problem:
This dialog is controlled internally. That means:
The parent component cannot control when it opens or closes You can’t reuse this dialog easily elsewhere without rewriting or duplicating the logic. It doesn’t follow the reusable component pattern in Vue 2 with Vuetify 2, which expects two-way binding using v-model.

✅ The Goal: Make the Dialog Reusable and Parent-Controlled Vuetify 2 v-model expects this:
You expose a prop called value.

You emit an event called input when the dialog should open or close.

This way, the parent component does:

<tip-dialog v-model="showDialog" :task="someTask" />

✅ Vue binds v-model="showDialog" to value prop and listens to @input events to update it.

### DIFF
--- a/src/components/atoms/TipButton.vue
+++ b/src/components/atoms/TipButton.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-dialog v-model="dialog" width="500">
+  <v-dialog :value="value" @input="$emit('input', $event)" width="500">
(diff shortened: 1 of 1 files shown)


marcgc21:
Closed as it exists another PR that fixes the problem

Issue #853, opened 2025-05-11
style: fix minor errors on user test


marcgc21:
Modify:

- [ ] Task title appears twice, with just one is enough
- [ ] Timer button must be hide, this is an information that the user should not see
- [ ] Tip button is ugly, it should be a button

<img width="1505" alt="Image" src="https://github.com/user-attachments/assets/52710e4c-d448-4dee-b6be-2820f375516b" />

PR #856, opened 2025-05-12
fixes style: fix minor errors on user test #853

saltykheera:
- Timer button is now Unclickable i.e user can see the timer icon button can not see the timer 
- why not remove the timer<TImer/> component serves a function and is integrated with the timer functionality removing it may cause integration issue for now this small tweek can help in short run
- just one primary title for all
image

- improved tip button
image


marcgc21:
@saltykheera the tip button needs to be a button with their own text or users do not understand what is it 🙏

marcgc21:
@saltykheera Did was solved with Juca, we can close it!

Issue #626, opened 2025-03-04
[🐞 BUG]:  Language Selection Not Working on Mobile


garvthakre:
### Description 📝

The functionality for changing languages is not working on mobile devices. While it works correctly on desktop, switching the language on a phone does not update the interface as expected.

### Link 🔗

[ /testlist](http://localhost:8080/testslist).

### Steps to Reproduce 🔄

1. Start the server using: npm run serve
2. Navigate to [/testlist](http://localhost:8080/testslist).
3. Attempt to change the language.
4. On a mobile device, observe that the language does not update.

### Screenshots 📸

![Image](https://github.com/user-attachments/assets/a55e7b5a-a122-401a-9ec9-807ebb6a10a5)

![Image](https://github.com/user-attachments/assets/4b863dbe-1613-4b5d-86ce-2c478f9a920d)

![Image](https://github.com/user-attachments/assets/c32f0927-d86d-47c8-a923-9f6dddd9eff4)

### Expected Behavior 🤔

![Image](https://github.com/user-attachments/assets/53bf2573-29d5-4b4a-b579-8b74fcaca993)

 The language changes successfully in PC/LAPTOP .

### Actual Behavior 😱

![Image](https://github.com/user-attachments/assets/4b863dbe-1613-4b5d-86ce-2c478f9a920d)

The language does not update in Phone.

### Environment 🌍

OS:Windows

subCode321:
will be glad to have a look at this. Can I be assigned?

PR #859, opened 2025-05-14
 Language Selection Not Working on Mobile  issue #626

rajputdivyanshu81:
Before :

image

After :

image

### DIFF
--- a/src/views/admin/DashboardView.vue
+++ b/src/views/admin/DashboardView.vue
@@ -113,9 +113,9 @@
 
             <v-spacer />
           </v-tabs>
-          <v-divider class="hidden-sm-and-down" />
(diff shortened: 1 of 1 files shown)


rajputdivyanshu81:
@marcgc21 sir can u review these changes

sahitya-chandra:
@rajputdivyanshu81 I dont think that this much amount of code should be changed, just to fix that translation on mobile, because I think it is working fine on large screen and other language.
Acc. to me adding new methods just for language translation is not efficient...

sahitya-chandra:
Iep, the second image shows it that for normal screen it is working perfectly...
May be you should check again if this really is the issue...

Issue #854, opened 2025-05-11
feat: current post question do not help


marcgc21:
After I finish a task, might be necessary to add a field where a link to some form. 
Right now the only option is doing that as a "Post question" which looks inneficient.

<img width="756" alt="Image" src="https://github.com/user-attachments/assets/f87a6020-9722-4037-8ef7-37b36c504103" />

rajputdivyanshu81:
@marcgc21   can u gave more context on this issue .

marcgc21:
When we create a new task we need an option under "Answer type" called: "Link form"
When it is selected it will open an input text field to place an url.

When we answer that task, this field must appear under the text: "Link to answer post-task form" 

<img width="1203" alt="Image" src="https://github.com/user-attachments/assets/6791af73-d3c3-4de9-b7e9-db3307ab5bb0" />

rajputdivyanshu81:
@marcgc21 can u assign me this issue

marcgc21:
Hi @rajputdivyanshu81, I might be awesome if you can finish it. Do you think that might be able to do it today? As you have seen we have pointed it as urgent!

I will be on touch during most of the day, if you have any question ask me and I will assist you.

rajputdivyanshu81:
@marcgc21 sir can u review PR_ID - #868

rajputdivyanshu81:
@marcgc21 sir ,i think this issue gets updated , u can close this issue

PR #868, opened 2025-05-16
Added link thing in the new task form

rajputdivyanshu81:
Resolve issue #854 ,

image

Added link to  the current tasks also -

image


marcgc21:
Hi @rajputdivyanshu81 take care because this input field already exists on develop branch, maybe you have an older version!
We need to add this new field here, it has to be a new option:
Screenshot 2025-05-16 at 18 00 16

PR #889, opened 2025-05-20
if a test has been answered it, it can not be possible to enter again…

rajputdivyanshu81:
Now if a user answered than, they cannot reanswered them , as in answer or in consent form also ..

image

Screenshot 2025-05-21 012736

Screenshot 2025-05-21 000635


marcgc21:
It has a direct impact with what we are doing on one of the GSOC projects. So we need to close it but thanks for you efforts!

PR #974, opened 2025-08-29
Div fix

rajputdivyanshu81:
Added div like in other container for consistency .


sahitya-chandra:
@rajputdivyanshu81, this file is no longer present in the codebase, cuz we have revamped the complete flow.

PR #984, opened 2025-09-26
Shows specific password validation requirements(min length/uppercase/…

### LINKED ISSUE #978: 🐛 Bug Report: Password validation inconsistency on Sign Up page
### Description 📝

On the Sign Up page, the password validation logic is inconsistent and shows misleading error messages.
• A password with 9 characters like Eshugupta still triggers the error:
“Password must be at least 8 characters”
• A password with 11 characters like Eshugupta01 also triggers the same incorrect error.
• A password like Eshugupta@01 (which contains special characters and numbers) is accepted.

This indicates that the error shown is not really about length but about complexity rules, yet the message only mentions length.

### Link 🔗

http://localhost:8080/signin

### Steps to Reproduce 🔄

1. Navigate to the Sign Up page.
2. Enter a valid email (e.g., kartikeyf12@gmail.com).
3. Try the following passwords:
• Eshugupta → Shows wrong error.
• Eshugupta01 → Still shows wrong error.
• Eshugupta@01 → Accepted.

### Screenshots 📸

Screenshot 1 – Password = Eshugupta (9 chars) → Incorrect length error
Screenshot 2 – Password = Eshugupta01 (11 chars) → Incorrect length error
Screenshot 3 – Password = Eshugupta@01 (accepted)

Image
Image
Image

### Expected Behavior 🤔

• If the password length is < 8, show:
“Password must be at least 8 characters.”
• If the password length is valid but it doesn’t meet complexity rules (e.g., missing number/special character), show a more specific message:
“Password must include at least one number and one special character.”

### Actual Behavior 😱

• The system always shows the length error (must be at least 8 characters), even when the password length is sufficient.
• The real rule (complexity requirement) is not communicated.

### Environment 🌍

• Browser: Chrome
• Environment: Localhost (emulator mode)
• OS: macOS

### Additional Information ℹ️

Possible Fix:
• Update password validation logic to separately check for length and complexity.
• Update error messages to clearly state the failed condition (e.g., missing uppercase, number, or special character).

Impact:
This bug creates confusion during onboarding  users with valid-length passwords may get blocked without understanding why.

AdityaShome:
## Summary
Updates the signup password validation to display the specific failing requirement (min length, uppercase, symbol) rather than always showing the generic “at least 8 characters” message.

## Why
Previously, any failure surfaced as a length error, misleading users when the real issue was missing uppercase or symbol.  
Improves UX and reduces signup friction.

## How it works
- `signupSchema.shape.password.safeParse(v)` is evaluated per keystroke.  
- On failure, the first issue’s message is shown, otherwise validation passes.

## Tested
- Password < 8 chars → shows “Password must be at least 8 characters”.  
- 8+ chars, no uppercase → shows “Must contain an uppercase letter”.  
- 8+ chars with uppercase, no symbol → shows “Must contain a symbol”.  
- Meets all requirements → form submits successfully.

### DIFF
--- a/src/views/public/SignUpView.vue
+++ b/src/views/public/SignUpView.vue
@@ -144,7 +144,12 @@ const emailRules = [
 
 const passwordRules = [
   v => !!v || t('errors.passwordRequired'),
-  v => signupSchema.shape.password.safeParse(v).success || t('errors.passwordValidate'),
+  v => {
+    const result = signupSchema.shape.password.safeParse(v)
+    if (result.success) return true
+    const firstIssue = result.error.issues?.[0]
+    return firstIssue?.message || t('errors.passwordValidate')
+  },
 ]
 
 const comparePassword = [


marcgc21:
Sorry @AdityaShome, this work was part of the GSOC problems that we were solving. Thanks for your time!

PR #1027, opened 2025-12-06
Fix: Implement Anonymous Authentication and Guest Access Flow #1010

### LINKED ISSUE #1010: fix: unknown user can not answer a user test

Santosh69:
### Summary:

Fixing the issue by initially signing in the unknown user (anonymously), hence providing the access to fetch the respective Test document from firestore (only works if user is logged-in), also storing the taskAnswers using nanoId.

***

**Flow (after fix):**

Test Link → `TestView.vue` → `Auth.js` → `AuthContoller.js` → `TestView.vue` → `Study.js` → `StudyController.js` → `Firestore` → `TestView.vue` → `UserTestView.vue`

https://github.com/user-attachments/assets/460bb5d1-a922-4097-940c-34550bce82c2

https://github.com/user-attachments/assets/3a03c60f-6a10-4914-baac-5eaca1e00e89

### **Root Cause:**

The reason why an unknown user wasn’t able to take the test is that there was no implementation of sign-in for the unknown user. The user needs to be signed-in to actually access Firestore (the DB), hence we were not able to access or use the study from Firestore, and it was silently failing (which was the minor bug) and not properly handled in the catch block. That’s why unknown users were not able to access those user tests (chicken and egg problem).

**Previous flow (before fix):**

Test Link → `TestView.vue` → Tries to fetch study → Firebase blocks the request (no access) → `getStudy` from `Study.js` returns `undefined` → no Test shown.

https://github.com/user-attachments/assets/5107e60f-eaf2-478e-8891-9567b8f65022

***

**Fix:**

### Changes in `TestView.vue`

- As soon as the user accesses `/testview` with `testId` and `token` (not required in case of public test), the user is signed in anonymously.  
- Only then the user is authorized to fetch the study with `testId`, and the study is assigned in the store as globally accessible (no namespace), so the user value is used everywhere.  
- Before actually loading the userTest through the child component `<UserTestView />`, the `testId` and `token` are validated using `validateStudyAccess` (from `studyValidation.js`), and the store action `getStudy` is fetched using `testId`.  

### Changes in `UserTestView.vue`

- Added an additional dialog to distinguish between logged-in users and unknown users, using `user.isAnonymous`.  
- After an unknown user takes the test, they are signed out and pushed to the sign-in page using `router.push('/signin');` (this route can be changed if needed).

### Changes in `Auth.js`

- Defined `signInAnonymously` action in the store, which gracefully calls `authController.signInAnonymously()` and handles/shows any errors gracefully.

### Changes in `AuthController.vue`

- Handling the real implementation of `signInAnonymously` .

### Changes in `AnswerController.vue`

- Slightly changed the `userDocId` generation logic for assigning the `TaskAnswer` respectively.  
- Previously it was sequential-index based (`Ev1`, `Ev2`, …) with \(O(N)\) lookup; now it uses `const tempId = nanoid(16)` which is safer, simpler, gives \(O(1)\) lookup, and allows storing the unknown user’s data even after the user is signed out.

https://github.com/user-attachments/assets/9e8253df-522c-4fef-adc9-f997077a077b

### Changes in `Study.js`

- Added a getter `isPublic` in the store, to understand whether the Test is public or not (used in `studyValidation.js` to appropriately provide access to each type).  

### Created `studyValidation.js`

- Handles study token validation for unknown users using `validateStudyAccess`, which is called in `TestView.vue` (the parent component).  

***

I’m glad to help with this and happy to keep contributing.
While working on this fix, I also noticed a couple of small bugs and edge cases that we can look at in future PRs if you want.
If you see any better way to handle the auth or validation logic, I’m totally open to your feedback and happy to update the PR.

Fix #1010

### DIFF
--- a/src/features/auth/controllers/AuthController.js
+++ b/src/features/auth/controllers/AuthController.js
@@ -8,7 +8,8 @@ import {
   sendPasswordResetEmail,
   setPersistence,
   browserLocalPersistence,
-  browserSessionPersistence
+  browserSessionPersistence,
+  signInAnonymously
 } from 'firebase/auth'
 import { auth } from '@/app/plugins/firebase'
 import axios from 'axios';
@@ -84,6 +85,10 @@ export default class AuthController {
       )
     })
   }
+  async signInAnonymously() {
+    console.log('Signing in anonymously')
+    return signInAnonymously(auth)
+  }
   // Reset Password
   async resetPassword(email) {
     // return sendPasswordResetEmail(auth, email)
@@ -94,22 +99,22 @@ export default class AuthController {
       template: 'passwordReset',
     })
   }
-
-  async autoSignIn() {
-    return new Promise((resolve, reject) => {
-      const unsubscribe = onAuthStateChanged(
-        auth,
-        (user) => {
-          unsubscribe()
-          resolve(user)
-        },
-        (error) => {
-          unsubscribe()
-          reject(error)
-        }
-      )
-    })
-  }
+  // I think this isn't needed(two SignIn)
+  // async autoSignIn() {
+  //   return new Promise((resolve, reject) => {
+  //     const unsubscribe = onAuthStateChanged(
+  //       auth,
+  //       (user) => {
+  //         unsubscribe()
+  //         resolve(user)
+  //       },
+  //       (error) => {
+  //         unsubscribe()
+  //         reject(error)
+  //       }
+  //     )
+  //   })
+  // }
 
   async deleteAuth(userId) {
     try {
--- a/src/features/auth/store/Auth.js
+++ b/src/features/auth/store/Auth.js
@@ -132,6 +132,21 @@ export default {
         throw err
       }
     },
+    async signInAnonymously({ commit }) {
+      try {
+        const { user } = await authController.signInAnonymously()
+        // Anonymous users don't have a userController profile yet, 
+        // so we just set the Firebase user to state
+        commit('SET_USER', user)
+      } catch (err) {
+        console.error(err)
+        commit('SET_TOAST', {
+          message: 'Error signing in anonymously',
+          type: 'error',
+        })
+        throw err
+      }
+    },
 
     async logout({ commit }) {
       try {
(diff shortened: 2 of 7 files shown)


Santosh69:
Hii👋🏻 @sahitya-chandra  , @marcgc21 , @KarinePistili 
I've submitted the fix for #1010 (Anonymous Authentication for Guest/Unknown Users).
Could you please review the changes? I'm happy to make any adjustments needed.

Thanks😀!

marcgc21:
Hi @Santosh69! Thank you for your efforts and detailed explanaiton about the process. I've not started reviewing the code but this issue looks very important because it cames from something that was broken during the last few months.

What you have implemented is okay, BUT! an unsigned or unknown user must be able to answer a test. How it works before the crash, if a user has no credentials into ruxailab he was able to answer the test and the results where stored under nanoId creation. Of course that session couldn't be recovered and if something during the process brokes the communication with the answer process we lose that answer.

As I understood, what you have implemented is that the user can start the process anonymously, however when he submits the answer he needs to fulfill the auth process??

Can you clarify this?

sahitya-chandra:
From the Description. This is not a correct solution to the issue. We want an anonymous user to be able to answer the user test without being authenticated, and most of its implementation is already done, but it got buggy somehow.

Santosh69:
Hiii @marcgc21 and @sahitya-chandra , thanks for the review and sorry about that😅, I think i quite misunderstood by the issue title. 

Yes, I agree that most of the implementation for this was already done but initially i noticed that firebase rules might also be cause of this issue, although changing them would only effect locally so, I switched to this approach ( user signs -in anonymously and takes the test and test is stored in the firestore and is signed out after submitting the answer succesfully).  

But I now understand that we want the unknown user to be able to answer the test without being authenticated ( in any manner)
and also we need to store their test answer, is that correct?

and does this issue have something to do with `firebase.rules`? 

I would still like to dig deeper(go through the code again) and debug and get followbacks with you, to find it, and fix it, if you certainly want me to, it would really be a help if you could maybe point out where to start?

Thank you

marcgc21:
> and also we need to store their test answer, is that correct?
Yep, we need to store it!

> and does this issue have something to do with firebase.rules?
in theory not

marcgc21:
I'm closing this PR because we already implemented the answer anonymous for the unmoderated test and also there is some kind of conflicts right now with the current branch and might be impossible to merge it.

PR #1042, opened 2025-12-09
fix: extra string quotes

DivyanshuVortex:


sahitya-chandra (approved):
Nice find @DivyanshuVortex

PR #1045, opened 2025-12-09
fix: resolve data loss on tab switch and concurrency overwrite (#1034)

### LINKED ISSUE #1034: [🐞 BUG]: Pre/Post-Test edits lost during editing
### Description 📝

When we create a **pre-test/post-tes**t variable for this user test, and save the variable, the moment they navigate between the edit test’s nav tabs(the data doesn’t persist), only when we save it using the "**Save**" button at the bottom, it does get saved

**Result**: user thinks data is saved, but it’s lost

### Link 🔗

https://ruxailab-prod.web.app/userTest/unmoderated/edit/(testId)

### Steps to Reproduce 🔄

1. Go to the Pre-Test tab.

2. Edit an existing question or add a new Pre-Test question.

3. Click into a question/variable title and modify it (normal editing).

4. Without explicitly saving(the bottom "Save" button, try to navigate away (for example, click another tab or section).

5. When the “Unsaved Changes” popup appears, click "Save"

### Screenshots 📸

https://github.com/user-attachments/assets/46298881-f6d8-4bc9-93a5-56613bb6bbd3

### Expected Behavior 🤔

Edits to Pre/Post-Test variables should always be preserved during normal editing. Temporary invalid states while typing should not cause user changes to be lost when saving or navigating away.

### Actual Behavior 😱

During editing, temporary invalid states prevent changes from syncing to the store. When the user clicks “Save,” an outdated state is saved and recent edits are lost.

### Environment 🌍

- Feature: Edit Test → Pre-Test / Post-Test
- Browser: Chrome (latest)
- User role: Logged-in user (test creator)

### Proposed Fix

- Allow `UserVariables.vue` to always sync the latest edits to the Vuex store, even when some variables are temporarily invalid (for example, during normal typing or when a new variable is created).

- Move responsibility for validation entirely to the parent `EditTestView.vue`, which already validates data before saving to the backend. This way, user edits are never lost, while invalid data is still prevented from being saved.

### Additional Notes

 I will go through this again and would like to work on this, open a PR (open to any pointers)
**Please assign this issue to me.**

Thanks!

Santosh69:
### Summary  

 As you were saying the admin conflicts when both edit the test (last change wins) so to handle this and ensure each change is preserved, used Snapshots, subscribed on the Mount and unsubcribed when Unmounted and then also in vuex store inside `EditTestView.vue`, the UI is also synced appropriately and the snapshot implementations is handled in `UserVariables.vue`  
 
 
 **Before fix:**

https://github.com/user-attachments/assets/1a4fcebb-68ea-420c-951b-c516d6a5acb5

**After Fix:**

https://github.com/user-attachments/assets/be29d979-2855-4fec-a76b-9f769bf265a8

###  Key Changes: 
 `EditTestView.vue`: Replaced the single-fetch logic with a Firestore `onSnapshot` listener. This ensures the Vuex store is always in sync with the latest DB state 
  `UserVariables.vue`: Added a `watch` to reactively update the local form whenever the Vuex store changes
  
  
  
### Steps to Reproduce 🔄

1. Go to the Pre/Post-Test tab.
2. Edit an existing question or add a new Pre/Post-Test question.
3. Click into a question/variable title and modify it (normal editing).
4. Without explicitly saving(the bottom "Save" button, try to navigate away (for example, click another tab or section).
5. When the “Unsaved Changes” popup appears, click "Save"

###   Additional Notes
 **Recommedation**:  We could add  `subscribeToTest` method to the Repository class (`FirebaseFirestoreRepository.js`) to keep it clean. 
 Let me know if you eed more context about the PR changes
 
 fixes #1034


Santosh69:
Hii👋🏻 @sahitya-chandra , @marcgc21 
I've submitted the fix for https://github.com/ruxailab/RUXAILAB/issues/1034 
Could you please review the changes? I'm happy to make any adjustments needed.

Thanks😀!

src/ux/UserTest/views/EditTestView.vue
KarinePistili, line 1: @Santosh69 could you change the onSnapshot logic to be placed on the StudyController? We want to place all db logic into controllers to follow our new code architecture.

KarinePistili (changes requested):
Thank you for the PR! Please take a look at the comment and tell us back when solved.

src/ux/UserTest/views/EditTestView.vue
Santosh69: sure @KarinePistili , have made the changes

Santosh69:
Hey @KarinePistili , I noticed the CI failures are coming from deprecated Vuetify props/events (left, outlined, @input/@change, etc.) in a few components.

This looks like a straightforward cleanup aligned with the current Vuetify API and should unblock the failing checks.
I can put up a small PR to fix these if that sounds good.
thanks!

KarinePistili (changes requested):
Structure looks okay, but there is no "StudySchema" model with a parse function which causes this error:

image

Please use the models we already have defined.

KarinePistili (approved):

Issue #1048, opened 2025-12-10
[Enhancement] : Improving the Error Handling in Auth.js


Nirvanjha2004:
Currently the Signup and signin page shows mostly the globalerrror :- " An error has occurred, sorry for the inconvenience " . I have fixed that to use proper errors coming from the firebase. For Example :-
If we signup with the same email again , then earlier it threw the globalerror , but now it throws the email exists error
Screenshots :- 

<img width="1919" height="1149" alt="Image" src="https://github.com/user-attachments/assets/c6a2d256-6cea-4506-9640-12452bbc7d69" />

<img width="1919" height="1005" alt="Image" src="https://github.com/user-attachments/assets/0ad61913-0629-4bc5-a77e-a0e4a7ac1c74" />

marcgc21:
What we should add here is a front validation in case of weak password, the rest we prefer to do not give much more information to the user.

Nirvanjha2004:
@marcgc21 Sure I will create a PR for this .

Nirvanjha2004:
Hi @marcgc21 , I checked the frontend and it seems there is already password validation present.
Currently it checks :-
1. Minimum 8 characters
2. At least one uppercase letter
3. At least one special character

Shall i add other validation checks as lowercase letters, numbers etc. ? or is this much fine.
Reference :- 

<img width="735" height="189" alt="Image" src="https://github.com/user-attachments/assets/491b4510-fd7b-4831-84e3-b6bf62747ae6" />

marcgc21:
I think that we already solve that. Going to close!

PR #1049, opened 2025-12-10
fix: Updated the error handling in Auth.js

Nirvanjha2004:
Issue #1048


marcgc21:
Hi @Nirvanjha2004 thank you for your efforts but we are going to close this issue because what it looks like an error indeed is correct. We don't want to provide so much information about if the user exists, if the password is true, if the email doesn't exists, etc. Showing. this kind of information is not a good practice. 

However, your issue brings some valid question which is the password weak validation that should happen on the front, instead of letting the server say that. So I will update the issue and closing that PR.

PR #1051, opened 2025-12-10
fix(auth): use .some() instead of .includes() and properly delete Firebase user

### LINKED ISSUE #1050: [🐞 BUG]: Using .includes() Caused Account Deletion Logic to Fail
### Description 📝

```js
```

As a result:

### Link 🔗

Account deletion page

### Steps to Reproduce 🔄

### Screenshots 📸

NA

### Expected Behavior 🤔

### Actual Behavior 😱

* Wrong delete path executed
* `user.delete()` never runs
* User Auth also remains

### Environment 🌍

### Additional Information ℹ️

_No response_

antaripdebgupta:
## Description
Correct provider check and ensure proper account deletion

## Related Issue
Fixes #1050

## Changes
* Replaced `.includes()` with `.some()` to correctly evaluate provider conditions:

```js
user.providerData.some(a => a.providerId !== 'google.com')
```

* Added:

```js
await user.delete()
```

to ensure the Firebase Auth account is fully deleted.

### DIFF
--- a/src/features/auth/views/ProfileView.vue
+++ b/src/features/auth/views/ProfileView.vue
@@ -863,7 +863,7 @@ const changePassword = async () => {
 const handlerDeleteConfirmText = async (value) => {
   const auth = getAuth();
   const user = auth.currentUser
-  if (user.providerData.includes(a => a.providerId !== 'google.com')) return deleteStep.value = 2
+  if (user.providerData.some(a => a.providerId !== 'google.com')) return deleteStep.value = 2
 
   try {
     isDeleting.value = true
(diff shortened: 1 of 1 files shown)


sahitya-chandra:
@antaripdebgupta, can you add a video of deletion working properly ??

antaripdebgupta:
@sahitya-chandra , 
## Before Fix
[beforeFix.webm](https://github.com/user-attachments/assets/c1f881ab-64bd-4bf6-aa9b-090ef618e9a0)

## After Fix
[afterFix.webm](https://github.com/user-attachments/assets/d73165d3-357d-4135-abfa-6cea75dd824b)

marcgc21:
The issue that tries to solve this PR needs to be changed because what happens that now is that ProfileView has some many logic that needs to came from controllers and not call directly firestore or auth operations.

PR #1061, opened 2025-12-12
simple fix to solve the issue: url updated, env variable updated

### LINKED ISSUE #1058: [🐞 BUG]: Missing SITE_URL config causes broken invite links + password reset redirects to localhost
### Description 📝

The Firebase Functions email system contains two related configuration issues:

1. **`SITE_URL` environment variable is used in `functions/src/https/email.js` but is never defined** in any `.env.example` or documentation.
2. **Password reset emails contain a hardcoded redirect URL (`http://localhost:8080/signin`)**, which breaks the flow in production.

### Link 🔗

https://ruxailab-prod.web.app/forgot-password

### Steps to Reproduce 🔄

1. Navigate to: https://ruxailab-prod.web.app/forgot-password
2. Enter a valid registered email.
3. Open the password reset email.
4. Click the reset link.
5. Enter a new password and click "Save".
6. Observe that Firebase redirects to: http://localhost:8080/signin instead of the production domain.

### Screenshots 📸

https://github.com/user-attachments/assets/e0586230-85e1-4081-93d3-1d2595c4a273

### Expected Behavior 🤔

- Password reset emails should generate: continueUrl = https://ruxailab-prod.web.app/signin
- The system should not contain any hardcoded localhost URLs.
- `.env.example` files should document this environment variables.

### Actual Behavior 😱

- `process.env.SITE_URL` is undefined
- Password reset emails include: continueUrl=http://localhost:8080/signin
- After resetting their password, users are redirected to localhost instead of the production site.

### Environment 🌍

- OS: Macos and ubuntu
- Browser: Firefox and Arc

### Additional Information ℹ️

The actual password reset works (handled by Firebase internally), but the post-reset redirect is broken because the URL is hardcoded. This results in a poor user experience and breaks expected production behavior.

Arka-Saha:
I solved the error with a very simple fix, env variable added, and url is updated so that localhost url redirect error won't come.

* Fixes #1058 

Hope this works, do let me know if any further improvements is needed.
Thank You!


marcgc21:
We are going to close this PR because their implementation is not good at all as the .env file that is modifying is not the correct env. However the idea is good, we are going to solve it!

Arka-Saha:
@marcgc21 thank you for the comment. I will definitely try to find a better implementation for the same. Thanks !

PR #1071, opened 2025-12-13
fix: complete missing translation keys to prevent language fallback

rahul810050:
### Solution

- Used en.json as the source of truth
- Updated all other language JSON files to include all required keys
- Ensured consistent structure and key coverage across all languages
- Prevented unintended fallback to English when selecting supported languages

### What changed

- Normalized all translation files under src/language/
- Added missing keys to non-English language JSONs
- Maintained existing translations while filling gaps where needed

### Demo Video
[Proper explanation](https://app.screencastify.com/watch/YM3CNa7JnHcHcrQvxm79)


rahul810050:
Hii @marcgc21 @sahitya-chandra !! could you please review it whenever you get a chance?

marcgc21:
This PR is very interesting but ends up to my mind with another question. It might be possible to just mantain the site with one languge and auto-translate the site authomatically like other sites?

rahul810050:
Hey @marcgc21  that’s a really good point!!

Auto-translation could definitely work well for content or marketing type pages, and I agree its an interesting direction. For this PR though the main goal was to fix the fallback issue and make the i18n setup predictable especially for UI strings where context, validation messages, and small wording differences really matter.

By using `en.json` as the source of truth and aligning all other languages to it, we avoid unexpected English fallbacks and keep the structure consistent. This also keeps the door open for future improvements including a hybrid or auto translation approach on top of a solid base.

Happy to adjust or explore alternative approach if that’s something we want to move toward

rahul810050:
@marcgc21 @jvJUCA could you please review this ???

Issue #1075, opened 2025-12-13
[🐞 BUG]: Export and Refresh buttons in Answers Timeline are non-functional


rahul810050:
### Description 📝

The Export and Refresh buttons in the Answers Timeline section appear clickable but do not perform any action.

The AnswersTimeline component correctly emits `export` and `refresh` events, however the parent component (`GeneralAnalysis.vue`) does not implement the corresponding event handlers. As a result, clicking these buttons provides no user feedback and does not trigger any functionality.

### Link 🔗

https://ruxailab-prod.web.app/userTest/unmoderated/answer/ answerId create a test and then  give the answer then go to answer page 

### Steps to Reproduce 🔄

1. Navigate to the Analytics / General Analysis page User answer page.
2. Scroll down to the **Answers Timeline** section.
3. Click on the "Export" button.
4. Click on the "Refresh" button.
5. Observe that no action occurs.

### Screenshots 📸

Proper Video explanation 
Link: https://app.screencastify.com/manage/videos/8QqhX1kRs2DMgH9QXB14?

### Expected Behavior 🤔

- Clicking "Export" should download the Answers Timeline data.
- Clicking "Refresh" should reprocess and update the timeline chart with the latest available data.
- The buttons should provide meaningful user interaction or feedback.

### Actual Behavior 😱

- Clicking "Export" does nothing.
- Clicking "Refresh" does nothing.
- No file is downloaded, no data refresh occurs, and no feedback is shown to the user.

### Environment 🌍

- OS: ArchLinux
- Browser: Chrome

### Additional Information ℹ️

The issue occurs because the parent component binds `@export` and `@refresh` events without defining the corresponding handler methods (`onExportTimeline`, `onRefreshTimeline`). This results in silent failures when the buttons are clicked.

A fix involves implementing these handlers or disabling the buttons until functionality is available.

rahul810050:
@marcgc21 @sahitya-chandra could you please review it ??

marcgc21:
Solved!

PR #1077, opened 2025-12-13
Fix non-functional Export and Refresh buttons in Answers Timeline

rahul810050:
### Problem

The Export and Refresh buttons in the Answers Timeline section were visible and clickable but did not perform any action.

Although `AnswersTimeline.vue` correctly emits export and refresh events, the parent component (`GeneralAnalysis.vue`) did not implement the corresponding event handlers. This resulted in a silent UX issue where users received no feedback after interacting with the buttons.

### Fixed issue
issue : #1075 

Solution

This PR implements the missing handlers in the parent component and ensures both actions work as expected:

- Export

  - Downloads a CSV file containing the same aggregated data shown in the timeline chart (last 30 days, grouped by date).
  - Respects current filters and search terms.

- Refresh

  - Forces the timeline chart to recompute and redraw using the latest available data.
  - Works correctly with reactive data and does not reload the page.

The fix is scoped only to the relevant components and does not affect other analytics sections.

### Demo Video
[Proper explanation](https://app.screencastify.com/watch/zdfOtE8Qb3oey6pXuOu2)

### DIFF
--- a/src/ux/UserTest/components/UnmoderatedTestAnalytics/GeneralAnalytics.vue
+++ b/src/ux/UserTest/components/UnmoderatedTestAnalytics/GeneralAnalytics.vue
@@ -940,6 +940,95 @@ onMounted(() => {
   // Create initial charts
   setTimeout(() => createTaskCharts(), 1000);
 });
+
+// Refresh timeline data
+const onRefreshTimeline = async () => {
+  try {
+    if (store.dispatch) {
+      const possibleActions = [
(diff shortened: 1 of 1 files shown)


marcgc21:
It looks okay despite this graphic it doesn't work very well and doesn't make much sense

rahul810050:
Thanks for the review and for merging this @marcgc21 !!
Noted your feedback on the graphic. happy to revisit or improve it in a follow-up PR if needed.

Issue #1079, opened 2025-12-13
[🐞 BUG]: Inconsistency in naming across the codebase


Nirvanjha2004:
### Description 📝

Currently there is inconsistency across codebase . somewhere the code is using answerDocId and somewhere it is using answersDocId which can lead to issues in future.

### Link 🔗

N/A

### Steps to Reproduce 🔄

1. Search for answerdocid and answersDocId 
2.  They both serve the same purpose and hold the same value

### Screenshots 📸

Look at the console log 
<img width="1919" height="845" alt="Image" src="https://github.com/user-attachments/assets/059861b6-5dce-4704-ac0f-1a4429e4196d" />

<img width="825" height="406" alt="Image" src="https://github.com/user-attachments/assets/74bf9249-9d16-4a5d-b3d0-f610c181cc39" />

### Expected Behavior 🤔

Only either of answersdocid or answerdocid should be used. 

### Actual Behavior 😱

we using both answersdocid and answerdocid

### Environment 🌍

Git branch :- Develop
OS - Windows 11

PR #1080, opened 2025-12-13
changed the answerDocId to answersDocId to remove inconsistency

Nirvanjha2004:
issue #1079


marcgc21:
It makes sense. I tested many options and everything seems working.

PR #1101, opened 2025-12-16
fix(security): implement zod schema validation for study data retrival

### LINKED ISSUE #1100: Fix missing validation for Study data API responses
### Description:

**Context:**
Currently, `StudyController.js` fetches data from Firestore and instantiates it directly via `instantiateStudyByType` without verifying if the data structure is valid.
We have `zod` installed in the project but it is currently unused in the source code.

**The Issue:**
If a database record is malformed (e.g. missing `testType` or `testTitle`), the application propagates the invalid object to the UI, leading to runtime white-screen crashes when components try to read undefined properties.

### Proposed Solution:
Implement a safety validation layer in `StudyController`:

1.  Import the existing `zod` library.
2.  Define a strict `StudySchema` for core fields (`id`, `testType`, `title`, `admin`).
3.  Use `.passthrough()` to allow non-breaking extra fields (avoiding a massive schema overhaul).
4.  Wrap the instantiation in a `try/catch` block to handle validation errors gracefully with proper logging instead of crashing.

Santosh69:
### Summary
I added a simple **Zod** check in the controller to validate study data before we use it. This stops the app from crashing if the database returns a broken or missing record, turning a white-screen error into a handled log.

### Addtional changes
**Note:** we could also split this out into a dedicated `schemas/` folder with a global validation helper, but that would be a much larger refactor (around ~250 lines). I stuck to this simpler version for now to fix the security hole quickly, but **I'm happy to switch to the full structure if you prefer that approach. i would like to get the approval before this change..**

### DIFF
--- a/src/controllers/StudyController.js
+++ b/src/controllers/StudyController.js
@@ -11,6 +11,21 @@ const COLLECTION = 'tests'
 const answerController = new AnswerController()
 const userController = new UserController()
 
+
+const StudySchema = z.object({
+  id: z.string(),
+
+  testType: z.enum(['USER', 'HEURISTIC', 'CARD_SORTING', 'MANUAL', 'AUTOMATIC']),
+  testTitle: z.string().min(1),
+  testAdmin: z.object({
+    email: z.string().email(),
+    userDocId: z.string()
+  }).passthrough(),
+  creationDate: z.number().optional(),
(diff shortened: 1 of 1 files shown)


sahitya-chandra:
Thanks for your contribution, but we currently don't need this...

PR #1104, opened 2025-12-17
refactor: remove duplicate function block.

amishhaa:
Same code block exists on lines:
https://github.com/ruxailab/RUXAILAB/blob/develop/src/features/auth/controllers/AuthController.js#L72
https://github.com/ruxailab/RUXAILAB/blob/develop/src/features/auth/controllers/AuthController.js#L98


sahitya-chandra (approved):
LGTM

PR #1111, opened 2025-12-19
fix: correct month name spellings

rakshityadav1868:
BEFORE


marcgc21:
It makes sense, eventually we need to delete spanish strings

Issue #1114, opened 2025-12-20
[🐞 BUG]: The Final Report stepper UI breaks after clicking the “Next” button.


Nirvanjha2004:
### Description 📝

When moving from Step 1 (Report Conclusion) to Step 2 (Generate Report), the stepper header updates but the content area becomes empty or incorrectly rendered.

This happens due to mismatched step values and incorrect step initialization in the Vuetify v-stepper implementation.

### Link 🔗

http://localhost:8080/heuristic/finalreport/

### Steps to Reproduce 🔄

1. Navigate to Final Report page.
2. Ensure heuristic answers exist (so the stepper view is rendered).
3. Click on the Next button in Step 1 – Report Conclusion.
4. Observe the UI behavior after the step change.

### Screenshots 📸

https://github.com/user-attachments/assets/cff7f13e-5c93-45b2-a4ff-8b4a48fc37c2
Before Fix

https://github.com/user-attachments/assets/c884dd33-6f50-487f-8e82-eebb67a5b9ad
After Fix

### Expected Behavior 🤔

Clicking Next should:
1. Save the conclusion text.
2. Move cleanly to Step 2 – Generate Report.
3. Display the corresponding content (FinalReportSelectionBox) correctly.

### Actual Behavior 😱

1. The stepper header advances to the next step.
2. The stepper content area appears blank or partially rendered.
3. UI looks broken or stuck between steps.

### Environment 🌍

Branch: fix/final-report-stepper
Framework: Vue 3 + Vuetify 3
Component: FinalReport.vue

PR #1115, opened 2025-12-20
fixed the Next button

Nirvanjha2004:
Root Cause:-
step was initialized with 0, while stepper values start from 1.
Stepper value props were strings ("1", "2") but step was a number.
:complete="value > 1" referenced an undefined variable (value).

Fix Summary:- 
Initialize step with 1.
Use numeric values consistently across v-stepper-item and v-stepper-window-item.
Replace invalid value references with step.

issue #1114

### DIFF
--- a/src/ux/Heuristic/views/FinalReportView.vue
+++ b/src/ux/Heuristic/views/FinalReportView.vue
@@ -1,109 +1,62 @@
 <template>
-  <PageWrapper
-    :title="answers.length > 0 ? $t('Final Report') : ''"
-    :loading="loading"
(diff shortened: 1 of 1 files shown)


sahitya-chandra (changes requested):
Can you add a before and after video of the changes

Nirvanjha2004:
before fix :- 

https://github.com/user-attachments/assets/8b3918a9-4250-4daa-97d7-660c30a502c4

after fix:- 

https://github.com/user-attachments/assets/5f91986f-3a9b-4c06-b380-0fb7cb8329c2

@sahitya-chandra

sahitya-chandra (approved):
LGTM

PR #1118, opened 2025-12-20
perf(iris-tracker): lazy load tensorflow to optimize initial bundle size

### LINKED ISSUE #1117: Performance: Lazy Load TensorFlow to Reduce Initial Bundle Size
### Problem

 The application currently imports entire  `tensorflow/tfjs `statically in `IrisTracker.vue`.
This causes the entire heavy TensorFlow library (approx. 20MB+) to be bundled and downloaded on every page load (including Sign In), even if the user never runs a test.

### Proposed Solution: 
Refactor `IrisTracker.vue` to use Dynamic Imports (`await import(...)`)for specific submodules (`tfjs-core`, `backend-webgl`) instead of the entire library. This ensures that TensorFlow dependencies are only fetched when the component is actually required by the user.

### Benefits

- Eliminate large TF chunks from the initial page load.
- The app feels snappier because the browser isn't choking on a massive file while trying to render the page

Santosh69:
### Fixes [#1117 ](https://github.com/ruxailab/RUXAILAB/issues/1117)

**Before Fix:**
**Flow**: `Router` -> `TestView` -> `UserTestView` -> `IrisTracker` -> TensorFlow (ALL TIED TOGETHER).

**After Fix:**
**Flow Broken:** `Router` -> `TestView` -> `UserTestView` -> `IrisTracker` 
Only when User Clicks "Start Test" we download and import TensorFlow
### Key Changes:
- **Performance**: Switched TensorFlow to lazy-loading, reducing the initial bundle size by ~20MB (downloads only when needed).
- **UX**: Implemented parallel loading for faster startup and added a Toast notification to handle network errors gracefully.

### DIFF
--- a/src/ux/UserTest/components/IrisTracker.vue
+++ b/src/ux/UserTest/components/IrisTracker.vue
@@ -4,11 +4,31 @@
 
 <script setup>
 import { ref, watch, onMounted, onBeforeUnmount, toRaw, computed } from 'vue'
-import * as tf from '@tensorflow/tfjs'
-import * as faceLandmarksDetection from '@tensorflow-models/face-landmarks-detection'
-import '@tensorflow/tfjs-backend-webgl'
 import { useStore } from 'vuex'
 import { getStorage, ref as storageRef, uploadBytes, getDownloadURL } from 'firebase/storage'
+
+let tf
+let faceLandmarksDetection
+
+const loadDependencies = async (store) => {
+    try {
+        [tf, faceLandmarksDetection] = await Promise.all([
+            import('@tensorflow/tfjs-core'),
+            import('@tensorflow-models/face-landmarks-detection'),
+            import('@tensorflow/tfjs-backend-webgl')
+        ])
+
+        return true
+    } catch (error) {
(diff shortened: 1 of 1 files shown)


Santosh69:
Hii @KarinePistili , @marcgc21, could you review this, whenever you  have time.
We could also remove `@tensorflow/tfjs` from `package.json`, if that’s okay

thank you.

Santosh69:
Hii @jvJUCA  @KarinePistili , do you want revert or apply any other changes for this issue.. is it valid?

marcgc21:
This is an interesting issue @jvJUCA. When I tested eye-tracker I found that some screen blocks and I guess that is because it is downloading this stuff. Can you review it?

jvJUCA (approved):
Congratulations! 🎉 Your pull request has been accepted! Thank you for your valuable contribution.

Your changes have been merged successfully and will be included in the develop branch. Your effort in improving our project is greatly appreciated.

Thanks again for your contribution, and we look forward to your continued involvement in our project!

Best regards,
@jvJUCA

PR #1126, opened 2025-12-21
Fix: Critical Privacy and db vulnerability in add cooperators.

amishhaa:
Currently when we attempt to add cooperators we are presented with a huge list to choose from, apart from reasons like privacy, it is not optimal for us to render thousands of users(and potentially lakhs as the db grows) everytime we render this page from the ruxailab users database.
From Ruxailab Prod  thousands of emails are exposed:
Screenshot 2025-12-21 at 6 20 43 PM

In this PR i have replaced the dropdown with a textbox and toast for invalid emails.

https://github.com/user-attachments/assets/1aa9bb49-01b6-4836-8544-4cc90a5081ca

### DIFF
--- a/src/shared/components/dialogs/InviteDialog.vue
+++ b/src/shared/components/dialogs/InviteDialog.vue
@@ -8,13 +8,9 @@
                 {{ title || 'Send Invitation' }}
             </v-card-title>
             <v-card-text class="pt-4">
-                <v-combobox :key="comboboxKey" ref="combobox" v-model="comboboxModel" :items="users.filter(user => user?.email != null)" item-title="email"
+                <v-combobox :key="comboboxKey" ref="combobox" v-model="comboboxModel"
                     :label="selectLabel || 'Select cooperator'" multiple variant="outlined" density="comfortable"
                     @update:model-value="validateEmail">
-                    <template #no-data>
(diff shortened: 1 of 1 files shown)


amishhaa:
@marcgc21 @KarinePistili Please share your ideas on this.

marcgc21:
Although we're exposing emails (which is something that we should advice in the privacy policy) the problem is that without the email we don't know how to add people as a collaborator.

I mean, we need some kind of identifyer that help us to find people if we want to add them into the study.

For now we'll close this PR and we'll think about what to do next.

amishhaa:
@marcgc21 
Great points!! one more thing to add is if we have all the people listed then someone can add collaborators who do not want to be added in the study, for example if i have an account in Ruxailab for attending study of xyz administrator, i would not liked to be added to study of other abc administrator unless i explicitly share my consent to attend the test by sharing my mail with the administrator which is my identifier, which was what i thought when i was thinking about reasons this was done like this because currently anyone can add anyone from the mail list even if they dont want to be a part of that study.
Also this can be easily abused if we have a huge db, by someone recurrently calling this page and we  recurrently fetching the entire db.
I am so glad i was able to bring this to your attention and contribute meaningfully, thank you!

PR #1131, opened 2025-12-21
Fix: fixed the “View All” button in Active Studies Overview

rakshityadav1868:
## Before:

https://github.com/user-attachments/assets/607c26a8-f415-45a2-bf8a-10d2c6b62ffa

## After:

https://github.com/user-attachments/assets/237b8425-e9bd-4c9a-a2d6-a1c390bc4d97


marcgc21:
It looks valid!

PR #1133, opened 2025-12-22
fix(bug): reset navigation guard state after successful save (#1132)

### LINKED ISSUE #1132: [🐞 BUG]: "Save and Leave" triggers infinite confirmation loop
## Summary

When a user selects "**Save and Leave**", the application successfully saves the data but fails to reset the internal "**unsaved changes**" flag. Result: The confirmation modal appears again erroneously on the immediate next page, trapping the user in a loop.

### Root Cause:
The global `dialogLeave` state remains `true` even after a successful save. The destination page detects this active flag on mount and triggers the modal again.

Santosh69:
### Summary: 
**The Fix:** Updated `SettingsView` to explicitly set `dialogLeave` to `false` after a successful save operation.

### Before Fix:

https://github.com/user-attachments/assets/9e18c4e1-031c-400e-b95c-af1368edf29c

### After fix:

https://github.com/user-attachments/assets/d8fd0747-52af-45b9-9647-7059de161eaa

### DIFF
--- a/src/shared/views/SettingsView.vue
+++ b/src/shared/views/SettingsView.vue
@@ -642,6 +642,7 @@ const validate = (valid, index) => {
 const onSubmit = async () => {
   await submit();
   store.commit('SET_LOCAL_CHANGES', false);
+  store.commit('SET_DIALOG_LEAVE', false);
   router.push({ name: store.state.pathTo });
 };
 


Santosh69:
Hii @KarinePistili @marcgc21  @sahitya-chandra , could you review this..

sahitya-chandra (approved):
LGTM

PR #1153, opened 2025-12-24
fix(profile): prevent focus on profile inputs when not editable

samarthdhoble:
### Problem
Profile input fields appeared editable even when the user had not enabled edit mode. The fields could receive focus and show a blinking cursor, but typing was not allowed, which resulted in a confusing UX.

### Root Cause
Vuetify `v-text-field` components were marked as `readonly`. In Vuetify, `readonly` inputs can still receive focus and display a cursor, even though their values cannot be modified.

### Solution
Replaced `readonly` with `disabled` for profile fields (username, email, contact number, and country) so they no longer accept focus when not editable. This aligns the UI behavior with user expectations and prevents misleading interaction.

### Notes
This is a minimal UI-only fix and aligns with the existing design where profile data is edited through the dedicated edit dialog.

### DIFF
--- a/src/features/auth/views/ProfileView.vue
+++ b/src/features/auth/views/ProfileView.vue
@@ -41,6 +41,7 @@
                   <v-icon start color="primary">mdi-account-details</v-icon>
                   {{ $t('profile.personalInfo') }}
                 </v-card-title>
+
                 <v-card-text>
                   <v-form>
                     <v-row dense>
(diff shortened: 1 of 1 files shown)


samarthdhoble:
This PR is ready for review. Happy to make any changes if needed.

marcgc21:
Thanks for the efforts but this PR is related to some refactor that needs to be done for the profile itself. Indeed the "Account" tab needs to be removed and a new button must appear under the left bar that allows to open the modal that edits this information. So I'm proceeding to close as we're looking for something different.

samarthdhoble:
Thanks a lot for the clarification and for explaining the roadmap.

That makes sense — I understand now that the Account tab will be removed and the edit flow will be moved under the left sidebar button.

I’ll keep an eye on the profile refactor and would be happy to help once the new structure is in place. Thanks again for the guidance!

PR #1113, opened 2025-12-19
fix: prevent date shifting by using local date formatting

rakshityadav1868:
fixes #1112 

This was because the date was being formatted due to `toISOString()` , as it convert the date to utc and timezone shiftes.
fixed this by using local date formatting.

https://github.com/user-attachments/assets/6af6454a-35da-4947-8cd6-9a96479b357f

PR #1162, opened 2025-12-26
fix: prevent date shifting by using local date formatting

rakshityadav1868:
This is a resubmission of #1113


marcgc21:
Awesome!

PR #1184, opened 2025-12-28
Fix: NASA-TLX Analytics tab not showing active underline when selected

rakshityadav1868:
When the user opens the NASA-TLX Analytics tab, the tab content loads correctly, but the active underline/highlight style is missing for this tab.
Other analytics tabs display the active underline properly, but NASA-TLX Analytics does not follow the same pattern. This makes the tab styling inconsistent across the analytics navigation.

BEFORE (3)

### DIFF
--- a/src/ux/UserTest/components/UserTestAnswer.vue
+++ b/src/ux/UserTest/components/UserTestAnswer.vue
@@ -16,32 +16,33 @@
             v-model="tab"
             bg-color="transparent"
             color="#FCA326"
+            slider-size="4"
           >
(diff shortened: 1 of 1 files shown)


src/ux/UserTest/components/UserTestAnswer.vue
```diff
@@ -16,32 +16,33 @@
             v-model="tab"
             bg-color="transparent"
             color="#FCA326"
+            slider-size="4"
```
marcgc21: Why this slider-size=4? What happens if we have more than 4 items?

PR #1196, opened 2025-12-29
Feat/update loading indicator v2

### LINKED ISSUE #1036: feat: update loading indicator
When we enter on places like "Settings" and the internet is on low mode, it can be possible to see a loading indicator on left top. We need to update this.

I would like to create a logo animation using the "X" that appears on loading and put it into the center of the screen.

revaarathore11:
Fixes https://github.com/ruxailab/RUXAILAB/issues/1036

This PR updates the global loading indicator described in Issue https://github.com/ruxailab/RUXAILAB/issues/1036, replacing the small top-left spinner with a centered, branded X-logo animation for a more polished and consistent user experience during page transitions.

BEFORE
image

AFTER
image

image

What was changed
• Full-screen overlay
• Centered animated X-logo
• Optional loading text
• Replaced the old circular loading indicator inside PageWrapper.vue with the new logo loader.
• Removed unused legacy loading import.
• Ensured loading behavior remains fully functional with existing loading and loadingText props.

Testing
• Verified that the new loader appears when navigating to pages like Settings, especially under slow network conditions.
• Checked responsiveness across screen sizes.
• Confirmed that the loader disappears correctly once content finishes loading.

📌Related Issue

Closes https://github.com/ruxailab/RUXAILAB/issues/1036


revaarathore11:
@marcgc21 @sahitya-chandra  please review

src/shared/components/Loading.vue
marcgc21, line 1: I think that there is no need to update that this file. Any specific reasons?

marcgc21:
We need to solve conflicts first 🙏

src/shared/components/Loading.vue
revaarathore11: @marcgc21 You’re right , it’s not strictly necessary to change this file.

The existing implementation would still work fine. I mainly updated Loading.vue to reduce duplication and keep all loading-related behavior in one place, since the same loader is now being used by PageWrapper and a couple of views.
I’m totally fine reverting this change and adjusting the PR.

revaarathore11:
@marcgc21 i resolved the conflicts .
please review!

revaarathore11:
@marcgc21 i reverted the changes in loading.vue 
and after that the loading logo does not appear anymore
so i think the changes i made are required for the feature
and they are safe to keep , as i tested it multiple times

revaarathore11:
@marcgc21 hey i used the red logo instead of the blue one as looked better
but i can change it to the blue one if you want just let me know once

marcgc21:
Not working as it was expected. Do not appear the logo neither on settings, cooperators, answers, etc.

Issue #1195, opened 2025-12-29
Fix: Non-Responsive Usability Test edit page


rahul810050:
Usability Test Edit page UI is non Responsive

Proper Video explanation [here](https://app.screencastify.com/watch/n1ejUWDBLqyUSMobAHn8)

PR #1198, opened 2025-12-29
Fix: Usability test edit page responsive

rahul810050:
Made Usability test edit page responsive
Fixed issue: #1195 

before Video : https://app.screencastify.com/watch/n1ejUWDBLqyUSMobAHn8

After Video: https://app.screencastify.com/watch/N84oAyz4B2NXB7eldDPo


marcgc21:
Hi @rahul810050, thanks for your efforts but this system needs a further understanding about how to display data which is hard to explain because we don't have it specified. I like the idea of the dropdown because i think that is similar about what you implemented for heuristic evaluation.

However I have few concerns, first related about margins and paddings which is something that I would like to remove in the cases of mobile and the second is that the dropdown vs tab system probably should be implemented to be modular and used by other modules. 

So as I have no clear definition about what to follow I will close this PR and I will take account to make it more responsive as it is right now.

rahul810050:
Hi @marcgc21 , thank you for the detailed feedback and for taking the time to review this PR.

I understand your concerns regarding margins/paddings on mobile and the need for a more modular approach for the dropdown vs tab system so it can be reused across modules. That makes a lot sense, especially given the lack of a clear specification.

Appreciate you considering the ideas here, and I’m happy to revisit or help once there’s a clearer direction or requirements defined. Thanks again for the guidance!

PR #1207, opened 2025-12-30
feat: export pre-test chart as PNG (#1202)

### LINKED ISSUE #1202: feat: export pre-test image as PNG
Add a three dot icon on the top right that allows to export this figure as PNG. 
Do not include legend. 
Will need to discuss if numbers must need to appear.

Image

Santosh69:
### Description: 
Added export functionality to SelectionPieChart component.

- Added v-menu with "Export as PNG" action.
- Uses `canvas.toDataURL()` to capture the chart.
- Inherently excludes the legend (which is outside the canvas).

### DIFF
--- a/src/ux/UserTest/components/answers/SelectionPieChart.vue
+++ b/src/ux/UserTest/components/answers/SelectionPieChart.vue
@@ -1,9 +1,26 @@
-<template>
-  <v-card class="pa-6 elevation-3 rounded-xl chart-card">
-    <div class="mb-4">
+<template> <v-card class="pa-6 elevation-3 rounded-xl chart-card">
+    <div class="d-flex justify-space-between align-start mb-4">
       <h4 class="font-weight-bold mb-2">
         {{ questionTitle }}
       </h4>
+      <v-menu location="bottom end">
+        <template v-slot:activator="{ props }">
+          <v-btn
+            icon="mdi-dots-vertical"
+            variant="text"
+            size="small"
+            v-bind="props"
+          />
+        </template>
+        <v-list density="compact">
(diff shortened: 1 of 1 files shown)


Santosh69:
Hii @marcgc21 @sahitya-chandra , could you review this and lmk if any changes needed

thanks!

marcgc21:
Hi @Santosh69 it looks okay however I think that we need to improve it a little bit. Use this as example:
Screenshot 2025-12-31 at 10 37 53
Screenshot 2025-12-31 at 10 38 11
Screenshot 2025-12-31 at 10 38 20

On those examples it appears the % of each value which it might be necessary to understand data values. Can you add this?

Santosh69:
Awesome, definitely @marcgc21 , 
I am thinking going ahead with this.. **because donut shape feels like good UI (haha) and the legends near each pie is a good UX, in my opinion ( will follow percentage)**
**also planning to add click based functionality which shows some details of each pie in the card** 

**References**
Gemini_Generated_Image_h74qtvh74qtvh74q
Gemini_Generated_Image_ajlwnnajlwnnajlw

marcgc21:
It might be an option but this maybe should be done into another issue. On this one the focus is the export feat. Achieving this, we could be able to use that funtionallity to export into papers.

Santosh69:
Hiii @marcgc21, could you review.. happy to make any changes needed..

thanks!

Santosh69:
i think we can improve a little.. @marcgc21

marcgc21:
It seems that what changed is the visualization on the front and not the pdf.
We're going to close it for looking for a different approach.
Screenshot 2026-01-07 at 13 45 27

Issue #1206, opened 2025-12-30
fix(user-test): fix timeline answers


marcgc21:
When we go into answers section on a user test, the timeline is not working well.
It doesn't make sense that the graphic starts on 30/11 if there is no data. So the graphic should start when we have data.

<img width="1184" height="598" alt="Image" src="https://github.com/user-attachments/assets/d8f75e1e-c613-401b-83fe-885a659cad3d" />

shubham220420:
Hey maintainers I want to work on this issue kindly assign me if possible:)

PR #1212, opened 2025-12-30
fix(user-test): fix timeline answers start date (#1206)

Santosh69:
### PR Summary

Fixed the Answers Timeline to dynamically start from the first valid answer date instead of a hardcoded 1-month lookback. This removes the misleading flat line at the start of the chart when data is recent. Logic gracefully falls back to the default window if no data is present.


marcgc21:
Hi @Santosh69, I'm going to close this PR because it is duplicated with #1214. That implementation looks better because it only fulfill the graphic with bullets only on the days that we have received answers.

PR #1228, opened 2025-12-30
refactor: implement reactive SUS & NASA-TLX satisfaction calculation

### LINKED ISSUE #1210: fix(user-test): satisfaction module on user test answers
Satisfaction module is still under "Coming soon" and not being calculated. There are many ways to calculate satisfaction on user tests, however as we use NASA-TLX or SUS we might be able to use that to fulfill that space.

In case that a test has a task that used SUS, show a SUS widget that displays satisfaction level.
In case that a test has a task that used NASA-TLX, show a NASA-TLX widget that displays satisfaction level.
In case that a test has a task that used SUS and NASA-TLX, show both metrics into the widget.
In case that the test has no task that uses SUS neither NASA-TLX, satisfaction level can not be calculated.

Image

rakshityadav1868:
fix #1210
Replaced the placeholder Satisfaction module with real calculations based on SUS and NASA-TLX answers.

This refactor introduces reactive computed metrics for SUS and NASA-TLX to avoid duplicated loops
and incorrect detection logic. The Satisfaction card now correctly displays a score whenever
relevant data exists and only shows the fallback message when neither SUS nor NASA-TLX data is available.


marcgc21:
To solve this issue it should be necessary to put the logic into the corresponding component and not into the "GeneralAnalytics" view.

Another thing, Satisfaction it doesn't need to be calculated as an score of 5. 

Your proposal:
Screenshot 2025-12-31 at 11 05 11

The current values that we have from that study:
Screenshot 2025-12-31 at 11 04 57

So in case of NASA-TLX you should use the value of "43.6" as not calculating a new value.

rakshityadav1868:
fixed @marcgc21

src/ux/UserTest/components/UnmoderatedTestAnalytics/SusAnalytics.vue
```diff
@@ -454,6 +454,17 @@ function getResponseLabel(score) {
     default: return 'Unknown'
   }
 }
+
```
marcgc21: I have doubts about this, in theory the 458 comment says that it is exporting the function. However there is no other instance of calling this. How it works this "defineExpose"??

rakshityadav1868:
Hi @marcgc21, I’ve removed the /5 rating and now show the native SUS and NASA-TLX averages. The logic is currently in GeneralAnalytics, but I can create `SatisfactionCalculator.js` utility and import it where needed. Does this approach look correct to you, or does this pr look good?

src/ux/UserTest/components/UnmoderatedTestAnalytics/SusAnalytics.vue
rakshityadav1868: yeah sorry my bad , this code is of no use i have removed it

marcgc21:
> Hi @marcgc21, I’ve removed the /5 rating and now show the native SUS and NASA-TLX averages. The logic is currently in GeneralAnalytics, but I can create `SatisfactionCalculator.js` utility and import it where needed. Does this approach look correct to you, or does this pr look good?

Yes, it makes sense!

rakshityadav1868:
Hey @marcgc21, I have changed my approach and fixed it

marcgc21 (approved):

PR #1188, opened 2025-12-28
feat #1187: added Storage module for centralized media management 

Santosh69:
### 🎯 Goal

"As a user I need a place to see all of the medias attached to all the studies that I have created... displayed into a table with the option to delete them."

### 🛠 Implementation Details
-   **Location**: Added "Storage" section strictly between **Templates** and **Notifications**.
-   **Visibility**: Created [StorageSection.vue](cci:7://file:///Users/sanhub/RUXAILAB/src/features/navigation/components/navbarSections/StorageSection.vue:0:0-0:0) to fetch and display **all** media files (video/audio/screen) from all user studies in a comprehensive table.
-   **Data Fetching**: Uses `AnswerController` to on-demand fetch answers, ensuring complete file visibility.
-   **Actionability**: Added "Delete" option (UI-mocked via alert) to allow users to manage storage costs.

### ✅ Verification
[View Fix](https://drive.google.com/file/d/1o_jQuNCzT-c414-g2cjM4iDeHxtXjARQ/view?usp=sharing)
[After UX Improvements](https://drive.google.com/file/d/1lJvQ6CfJwPvjGsG_zJFQyFB5Uitg1qvX/view?usp=sharing)

### 📝 Additional Notes
-   **Refactor Opportunity**: [formatBytes](cci:1://file:///Users/sanhub/RUXAILAB/src/features/navigation/components/navbarSections/StorageSection.vue:201:0-208:1) is implemented locally to keep changes isolated. Consider extracting to a shared `fileUtils.js` later to avoid duplication with `StorageInfo.vue`.

PR #1237, opened 2025-12-31
feat #1187: storage backend integration & UX improvements

### LINKED ISSUE #1187: feat: new module called "Storage"
As a user I need a place to see all of the medias attached to all the studies that I have created. 
They must be displayed into a table with the option to delete them.
This module needs to appear between "Templates" and "Notifications".

The idea behing that is to avoid increasing storage numbers as it can be expensive for those users that are using the platform and create many studies. In a near future we should limit this on the production environment because we can not afford users with 20Gb of information right now.

Santosh69:
## Summary
Added secure file deletion and real-time size fetching to the backend, the backend code now lives in its own dedicated 
`storage.js` module
### Backend (Secure Storage)
- [x] **Secure Deletion:** `deleteMediaFile` ensures strict ownership validation before deletion.
- [x] **Metadata:** `getFileMetadata` fetches real file sizes dynamically.
## Additional Notes
1. **Testing Proposal:** Should we add Unit Tests for this new backend logic in a follow-up PR?
2. **Delete Strategy:** Implements a "File-Only Delete" strategy (removes physical file to save costs, but preserves the Firestore record to prevent accidental metadata/note loss).
3. **Commit Structure:** I intentionally made the previous commits follow here, for you to separately test the frontend and backend part
### Future Work
**This PR updates the Storage Page with real data. The Dashboard Widget still uses mock data, should I update that here or in a separate PR?**
 
 Part of [#1188](https://github.com/ruxailab/RUXAILAB/pull/1188)


Santosh69:
Hiii @marcgc21 , could you review this.. happy to make any changes or reverts of the Additional notes

marcgc21:
@KarinePistili I'm worried about this part:

> Delete Strategy: Implements a "File-Only Delete" strategy (removes physical file to save costs, but preserves the Firestore record to prevent accidental metadata/note loss).

In theory if for some reason, we delete the file, it shouldn't be removed from the firestore also? I'm trying to think a way in which we wouldn't like to remove it but I can't find it. 

Any idea?

Santosh69:
**yeah  @marcgc21  felt exactly the same that is why I added as note, because the files might appear when we refresh (sometimes) , but when clicked would show error and not found etc..**

**we can make change to handle this, Instead of deleting the record, we could Nullify the link.**

Santosh69:
also lmk **if you have thoughts on unit tests and dashboard's mock data..**

Santosh69:
hii @marcgc21  @KarinePistili , could you review this. and also lmk your thoughts on comments?, before further implementation and to fully test and showcase

PR #1260, opened 2026-01-02
Feat: Add clickable step navigation in test progress bar

### LINKED ISSUE #1185: [✨ FEATURE]: Add Back / Review button on Final Submission screen
### Feature Description 📝

Add a (Back) button on the Final Message / Submit screen.
This button should take the user back to the previous step so they can review or edit their answers before submitting.

### Motivation 🌟

Right now, once users reach the final screen, they only see the Submit button. If they want to quickly check or fix any answer, there is no clear way to go back. Adding a Back / Review option will prevent accidental submissions and give users more confidence before completing the test.

### Expected Behavior 🤔

a) The final submission screen should show two buttons:
	       1) ← Review answers
	       2) Submit
	b) When the user clicks Review answers, they are taken back to the previous step.
	c) All previously filled answers should remain saved.
	d) The user should be able to edit their responses and then return to the final screen to submit.

### Additional Information ℹ️

Image
here in left side of submit button we can have a back /review btn.

rakshityadav1868:
fixes #1185 
## Description
Enable users to navigate back to any completed step in the test progress bar to review and modify their answers without losing form state

https://github.com/user-attachments/assets/6f58ac7c-c870-4ba2-9a7f-82f9ac463e72

### DIFF
--- a/src/ux/UserTest/views/ModeratedTestView.vue
+++ b/src/ux/UserTest/views/ModeratedTestView.vue
@@ -131,20 +131,75 @@
                 :class="{ 'stepper-animate': globalIndex === 4 && test?.testStructure?.userTasks?.length > 1 }"
                 style="visibility:visible">
                 <v-stepper-header>
-                  <v-stepper-item :value="1" title="Consent" :complete="stepperValue > 1"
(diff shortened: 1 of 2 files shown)


marcgc21:
Explained into the issue that this is not vaflid.

PR #1261, opened 2026-01-02
Fix: handle NaN values when no answers are provided

rakshityadav1868:
Prevents showing incorrect Most/Least Stressful dimensions when no NASA-TLX responses exist.
BEFORE (4)


marcgc21:
It seems it doesn't work. I created a blank test, with one task and added NASA-TLX type. If not answer the form, on "Answers" tab it appears the NaN information. I'm closing the PR and going to fix it. Thanks for finding the bug!

PR #1287, opened 2026-01-05
fix: Improve mobile dashboard layout and spacing

rakshityadav1868:
This PR improves the mobile dashboard layout for small screens.

It fixes spacing and alignment issues, ensures the top header and cards are displayed consistently, and moves the Recent Activity section to a more natural position in the mobile flow. 

These changes only affect the mobile layout and do not impact desktop view.

BEFORE (5)


marcgc21:
Although this PR is necessary, we don't recommend the intense usage of new CSS code as most of the things can be done using the classes that vuetify provide. Indeed, all the responsiveness issues are difficult to be addressed for newcomers because we're aware of many problems that are hard to be explained and we do not explained. So I'm going to close it and I will personally work on that sooner or later during this month.

PR #1301, opened 2026-01-07
fix(auth): move account deletion logic from ProfileView to controller and ensure proper deletion from firebase.

### LINKED ISSUE #1050: [🐞 BUG]: Using .includes() Caused Account Deletion Logic to Fail

rakshityadav1868:
fixes #1050 

https://github.com/user-attachments/assets/8ef5a53e-8bfc-49bd-879b-9752369bf2f3

### DIFF
--- a/src/features/auth/controllers/AuthController.js
+++ b/src/features/auth/controllers/AuthController.js
@@ -8,11 +8,15 @@ import {
   sendPasswordResetEmail,
   setPersistence,
   browserLocalPersistence,
-  browserSessionPersistence
+  browserSessionPersistence,
+  reauthenticateWithCredential,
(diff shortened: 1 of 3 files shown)


marcgc21 (commented):
We don't wanat to provide too much information when signinwithemail. Eve more, take care because the pull request is about proper deletion, not dealing with signin.

PR #1304, opened 2026-01-08
Fixed and move account deletion logic from view to controller

### LINKED ISSUE #1050: [🐞 BUG]: Using .includes() Caused Account Deletion Logic to Fail

rakshityadav1868:
fixes #1050 
resubmission #1301 
## Summary
 Logic is moved from ProfileView.vue to AuthController for better maintainability, testability, and reusability. Users can delete their accounts safely by verifying their identity through either email/password or Google authentication.
### AuthController.js
**New Methods Added:**
- **`deleteUserAccount(user, password)`**: Core deletion logic
  - Detects provider type (Google or email/password)
  - Reauthenticates user based on their sign-in method
    - Email/password users: Uses password credential
    - Google users: Shows Google OAuth popup
  - Deletes Firebase Auth user account
  - Calls backend cleanup endpoint
  - Automatically signs out user after deletion
  
- **`deleteAuth(userId)`**: Backend cleanup endpoint call
  - Posts to Firebase Cloud Function to clean up Firestore and storage
  - Logs errors without throwing (non-critical step)
### ProfileView.vue
- **Step 1 - Confirmation**:
  - `handlerDeleteConfirmText()`: 
    - Validates user is still logged in
    - Detects if user has email/password provider (needed for password field)
    - Moves to Step 2 of deletion flow
- **Step 2 - Authentication**:
  - Shows password field only for email/password users
  - Shows Google auth message for Google-only users
  - `handlerDeleteAccount()`:
    - Gets password (if email user) or null (if Google user)
    - Calls `authController.deleteUserAccount()` with appropriate credentials
    - Redirects to home page on success

https://github.com/user-attachments/assets/af8dc3e8-d285-4790-9771-9c70ee39bc67


KarinePistili (changes requested):
I understood the idea and I agree that is good to move as much logic as possible to the controller. The only improving point that needs to be done, is that right now it is duplicating code instead of improving what we have. We already had the action "deleteAuth" directing to the store and to a controller.  Please adjust to not have code duplicated and to keep using the store to manage the state.

rakshityadav1868:
Hi @KarinePistili , I have updated the implementation. The ProfileView now only dispatches the store action, and the store remains responsible for state management (SET_USER, toast, signOut).

KarinePistili (approved):
It worked. There was only an import error, but I fixed and I will approve it. Thank you for the PR!

rakshityadav1868:
Thanks @KarinePistili, I will keep this in mind for the future.

PR #1233, opened 2025-12-30
feat: implement sart instrument

rahul810050:
Fixed Issue: #1211

Video Explanation: https://app.screencastify.com/watch/OMyACxMNXrvO63ta9u6G

PR #1321, opened 2026-01-10
feat: implement SART instrument and analytics dashboard visualization #1211

### LINKED ISSUE #1211: feat: implement SART instrument
In the same way that we have SUS or NASA-TLX we would like to implement another form instrument which is called SART.

To solve this issue the user must be able to use the new instrument when he creates a task.
Also add a new view to display data after using that form such as SUS or NASA-TLX.

Golixco:
I worked on adding the SART (Situational Awareness Rating Technique) to the platform. It is now fully ready to use in the task creator, and I’ve added a section in the analytics dashboard so researchers can actually see the calculated scores.

**What I did:**

* **Added SART to the task selection:** You can now pick "SART" when creating a task.
* **Built the participant form:** Created the `SartStep.vue` component with the 1-7 sliders for Demand, Supply, and Understanding.
* **Added Analytics:** Created a card in the results view that automatically calculates the SA score using the Understanding - (Demand - Supply) formula.
* **Cleaned up the code:** Fixed the linting errors and updated the components to use Vuetify 3 standards (like `text-primary` and `size="large"`).

I've tested this locally and the math works perfectly (for example, a score of 7, 2, and 5 results in a 10.00). All my modified files pass `npm run lint` with 0 errors

Please let me know if you’d like any changes or refinements. Happy to iterate.

**Fixes: #1211**

### DIFF
--- a/src/app/plugins/locales/en.json
+++ b/src/app/plugins/locales/en.json
@@ -180,6 +180,7 @@
     "overallAnalyticsDescription": "View the aggregated analytics for your test.",
     "individualAnalytics": "Individual Analytics",
     "nasaTlxAnalytics": "NASA-TLX Analytics",
+    "sartAnalytics": "SART Analytics",
     "finalReport": "Add additional data to your test.",
     "intro": {
       "reports": "Keep track of your evaluators progress and answering status while they complete your test.",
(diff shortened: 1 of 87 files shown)


rakshityadav1868:
Hey @Golixco u may close this pr as it already implemented #1233

PR #1351, opened 2026-01-15
chore(deps): replace invalid UTC with Etc/UTC in dependabot config

antaripdebgupta:
Fix Dependabot config by using a valid IANA timezone (`Etc/UTC`) to resolve validation errors.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
       interval: 'weekly'
       day: 'monday'
       time: '09:00'
-      timezone: 'UTC'
+      timezone: 'Etc/UTC'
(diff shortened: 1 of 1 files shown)


sahitya-chandra (commented):
@antaripdebgupta Can you fix the failing CI check

antaripdebgupta:
Hi @sahitya-chandra,

Failure is from i18n Diff Guard, not the Dependabot change. This PR only updates the timezone value.

i18n Diff Guard issue is due to the PR checkout context and is already fixed in #1352. Once that is merged, this check should pass here as well.

marcgc21 (approved):

PR #1352, opened 2026-01-16
ci: fix i18n diff guard checkout context in PR workflow

antaripdebgupta:
Fixes CI failure where i18n-diff-guard.js was not found during PR runs.
Removes explicit checkout ref to rely on default PR checkout behavior.
No functional changes to the validator.

### DIFF
--- a/.github/workflows/i18n-diff-guard.yml
+++ b/.github/workflows/i18n-diff-guard.yml
@@ -15,7 +15,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Fetch base branch
         run: git fetch origin ${{ github.event.pull_request.base.ref }}


sahitya-chandra (commented):
@antaripdebgupta why is this check passing for all other prs?

antaripdebgupta:
Hi @sahitya-chandra,

This only failed when actions/checkout explicitly set a ref in PR context, which can check out the base branch instead of the PR head—so the script wasn’t present.
Other PRs pass because they use the default PR checkout behavior. Removing the explicit ref makes this workflow consistent with those cases.

marcgc21:
> Hi @sahitya-chandra,
> 
> This only failed when actions/checkout explicitly set a ref in PR context, which can check out the base branch instead of the PR head—so the script wasn’t present. Other PRs pass because they use the default PR checkout behavior. Removing the explicit ref makes this workflow consistent with those cases.

So this PR must be accepted? @antaripdebgupta

antaripdebgupta:
@marcgc21,
Yes — this fixes a PR-only CI issue caused by explicitly setting ref in actions/checkout, which can check out the base branch and miss the script. Other PRs pass because they use the default checkout behavior.

marcgc21 (approved):

PR #1379, opened 2026-01-21
Fixes heuristic color when all option values are equal

### LINKED ISSUE #1377: [🐞 BUG]: getColor returns amber/orange instead of green when max === min (Heuristic Test Answers)
### Description 📝

When `max === min`, the `getColor(value, max, min)` function never returns `green`
even when `value === max`.

This results in incorrect color representation in the Heuristic UI, where values
that should be marked as green (highest score) appear as amber/orange instead.

The issue occurs because the threshold calculation results in `h = 0`, causing the first threshold condition `(amber)` to match before the `green` case can be reached.

### Link 🔗

https://ruxailab-prod.web.app/heuristic/answer/mXSqvUFqgdYyVyXXFCEi

### Steps to Reproduce 🔄

1. Open a completed heuristic test evaluation or create a new heuristic test.
2. If creating a new test:
   - Go to the **Edit** tab.
   - Add **two heuristics**.
   - In the **Options** tab, add **two options**.
3. Set the value of each option to **1**, so that `max === min === value`.
4. Go to **Preview**, start the test, and complete it.
5. Optionally, use another user ID and repeat the test to generate multiple responses.
6. Go to the **Answers** tab, then select the **Heuristics** tab → **Answers by Evaluator**.
7. Observe that heuristic scores are displayed in **amber/orange** instead of **green**, even though they are the maximum value.

### Screenshots 📸

All heuristic values are equal **(value = max = min = 1)**, but getColor renders them as **amber/orange** instead of **green** in the **Heuristics → Answers by Evaluator section.**

- See attached screen recording demonstrating the issue.

https://github.com/user-attachments/assets/b401bf69-b33d-49a7-8988-d1a9580d785b

### Expected Behavior 🤔

When value === max and max === min, the heuristic value should be displayed in green,
as it represents the highest possible score.

### Actual Behavior 😱

When max === min, heuristic values are displayed in amber/orange instead of green,
even when value === max.

### Environment 🌍

- Vue component: src/ux/Heuristic/components/HeuristicTestAnswer.vue (color rendering logic using getColor)
- Function: getColor(value, max, min)
- Git branch: develop
- OS: Windows 11
- Browser: Chrome
- Firebase: Emulator + Production

### Additional Information ℹ️

Relevant function causing the issue:

```js
const getColor = (value, max, min) => {
  max = Number(max) || 0;
  min = Number(min) || 0;
  const h = max ? (max - min) / max : 0;

  if (value == null) return 'grey';
  else if (value === 0) return 'red';
  else if (value <= min + 1 * h) return 'amber';
  else if (value <= min + 2 * h) return 'orange lighten-1';
  else if (value <= min + 3 * h) return 'lime';
  else return 'green';
};
```

### Root cause:
When max === min, h = (max - min) / max = 0.
This makes the condition `value <= min + 1 * h` true, returning 'amber' before
the 'green' branch is reached.

Ady0333:
### **Description :**
Fixes: #1377 
Fixes an edge case in heuristic color calculation where results were incorrectly shown as amber when all option values were equal. In this scenario, the value represents the maximum possible score and should be displayed as green.

### **Impact :**
- Prevents misleading visual feedback in Heuristic Evaluation results 
- Ensures correct color classification when option ranges collapse (max === min)

### **Root Cause :**

When max === min, the threshold calculation evaluates to zero, causing the first range check to always match and return amber.

`h = (max - min) / max; // evaluates to 0 when max === min`

### **What has been fixed :**

Explicitly handle the equal-range case before applying threshold logic:

`if (max === min) return 'green';`

### DIFF
--- a/src/ux/Heuristic/components/HeuristicsTestAnswer.vue
+++ b/src/ux/Heuristic/components/HeuristicsTestAnswer.vue
@@ -653,15 +653,20 @@ const checkIfNan = (value) => {
 const getColor = (value, max, min) => {
   max = Number(max) || 0;
   min = Number(min) || 0;
-  const h = max ? (max - min) / max : 0;
 
   if (value == null) return 'grey';
-  else if (value === 0) return 'red';
-  else if (value <= min + 1 * h) return 'amber';
+  if (value === 0) return 'red';
+
+  if (max === min) return 'green';
+
+  const h = (max - min) / max;
+  
+  if (value <= min + 1 * h) return 'amber';
   else if (value <= min + 2 * h) return 'orange lighten-1';
   else if (value <= min + 3 * h) return 'lime';
   else return 'green';
 };
+
 
 const getColorPorcentage = (value) => {
   value = Number(value) || 0;


Ady0333:
Hello @sahitya-chandra and @marcgc21, Greetings!!
This PR fixes the issue no. 1377. Happy get feedback from you guys and make any changes if required. Just review it once.

Thankyou

marcgc21:
Hi @Ady0333 it looks interesting, however we require screenshots as a validation tool also. Can you add them please??

Ady0333:
Hey @marcgc21 , 
You are right, I must upload a screenshot of showing it running it locally. Allow me some couple of minutes I will upload a screenshot shortly.

chanchal430:
Nice work @Ady0333. I also worked on the same issue in my PR (#1382 ) with a slightly different minimal fix. Happy to align if @marcgc21 prefer one approach — great to see this getting addressed!

Ady0333:
Hey @marcgc21, i am having some trouble in locally runnung the server again. But earlier after fixing the issue I have verifies by running it locally. For now I am uploading this screenshots. Hope they find helpful !!

image

image

marcgc21:
Hi @Ady0333 it seems that this problem was already solved. Going to close 🙏

PR #1381, opened 2026-01-21
added username on signup

### LINKED ISSUE #1380: [✨ FEATURE]: Add optional username field during signup
### Feature Description 📝

Currently, when users sign up, only email and password are collected. The dashboard displays 'User' as the default username.Add an optional username field to the signup form so users can set their display name during registration.

Benefits:
Better user experience with personalized greeting on dashboard
Users don't have to go to profile settings to set their name after signup

### Motivation 🌟

Currently, when a new user signs up on RUXAILAB, only their email and password are collected. After registration, the dashboard greets them with a generic "Welcome back, User! 👋" message. This feels impersonal and requires users to navigate to their profile settings to set a display name.

By collecting an optional username during signup, we can provide a more personalized experience from the very first moment users access their dashboard.

### Expected Behavior 🤔

Signup Form
A new optional username field appears between the email and password fields
The field label shows "Username (Optional)" to indicate it's not required
If the user enters a username, it must be at least 3 characters long
Validation error appears below the field if username is 1-2 characters: "Username must be at least 3 characters"
Users can leave the field empty and still sign up successfully

After Signup
If username provided: Dashboard shows "Welcome back, [Username]! 👋"
If username empty: Dashboard shows "Welcome back, User! 👋" (current default behavior)
The username is stored in the user's Firestore document and displayed in the user menu

### Additional Information ℹ️

_No response_

ASHLESHA-dot:
This PR adds an optional Username field to the signup flow and stores it in the user profile document. If provided, the dashboard greeting uses the username; if left blank, the app continues to fall back to the default.

Newly created accounts currently show a generic dashboard greeting (e.g., “Welcome back, User!”). Asking for an optional username during signup improves personalization without blocking registration.

Changes
Added an optional username input on the signup form.
Added client-side validation: if username is entered, it must be ≥ 3 characters.
Passed the username through the signup action and persisted it in the Firestore user document.
Added/updated i18n strings for "Optional" label and the username min-length validation message.

https://github.com/user-attachments/assets/42297e40-c1c0-4e60-b785-ddc852a07c59

Fixes #1380

### DIFF
--- a/src/app/plugins/locales/en.json
+++ b/src/app/plugins/locales/en.json
@@ -221,7 +221,8 @@
     "authenticationCancelled": "Sign-in was cancelled. Please try again.",
     "networkError": "Network error. Please check your internet connection and try again.",
     "popupBlocked": "Your browser blocked the sign-in popup. Please allow popups for this site and try again.",
-    "tooManyAttempts": "Too many attempts. Please try again later."
+    "tooManyAttempts": "Too many attempts. Please try again later.",
+    "usernameMinLength": "Username must be at least 3 characters"
   },
   "Introduction": {
     "title": "UX Remote LAB",
@@ -766,6 +767,7 @@
     "previous": "Previous",
     "seeCode": "See the code",
     "username": "Username",
+    "optional": "Optional",
     "goToConsole": "Go to Console",
     "signout": "Sign-out",
     "profile": "profile",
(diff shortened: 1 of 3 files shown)


marcgc21:
Althought this is interesting we don't need that users add their username right now. We're going to close it for now.

PR #1382, opened 2026-01-21
fix(heuristics): Fix getColor returning amber instead of green when max === min

### LINKED ISSUE #1377: [🐞 BUG]: getColor returns amber/orange instead of green when max === min (Heuristic Test Answers)

chanchal430:
### Summary
Fixes an issue where heuristic values were displayed as **amber/orange** instead of
**green** when `max === min` and `value === max`.

### Root Cause
When `max === min`, the threshold calculation produced `h = 0`, causing the first
threshold condition to match before the `green` case could be reached.

### Fix
- Added an explicit guard for the `max === min` case.
- Ensured that values equal to `max` resolve to **green**.
- Preserved existing color behavior for all other scenarios (`red`, `amber`, `orange`, `lime`).

### Steps to Verify
1. Create or open a completed heuristic test.
2. Ensure all option values are identical (e.g., all values = 1).
3. Navigate to:
   - **Answers → Heuristics → Answers by Evaluator**
4. Confirm that heuristic scores are now rendered in **green**.

#### Screenshot (after fix): When all heuristic values are identical `(value = max = min = 1)`, the heuristic scores are now correctly rendered in green, indicating the highest possible score.
Screenshot 2026-01-21 192533

#### Screen recording (after fix): End-to-end validation showing that the `getColor` logic correctly handles the `max === min` edge case and displays the highest heuristic score in `green`.

https://github.com/user-attachments/assets/f2fbca05-a580-4bdb-bc03-3ec2d18b9250

### Notes
- No UI or behavioral changes outside the color-mapping logic.
- Backward-compatible with existing heuristic scoring behavior.


chanchal430:
Hi @sahitya-chandra @marcgc21 , I’ve updated the PR with the minimal fix for the `max === min` case and cleaned up duplicate checks. Could you please review when you get a chance? I’ve also added screenshots/tests for validation. Thanks!

src/ux/Heuristic/components/HeuristicsTestAnswer.vue
KarinePistili, line 1: variable "val" is not defined.
image

KarinePistili (changes requested):

chanchal430:
Hi @KarinePistili, Addressed the issue in the latest commit by correcting the range calculation and handling numeric edge cases. Colors now map correctly relative to min–max values.
Screenshot 2026-01-23 220752
Screenshot 2026-01-23 221427
Screenshot 2026-01-23 221702

KarinePistili (approved):

Issue #1401, opened 2026-01-22
Fix: TemplateInfoDialog console warning on empty templates


rahul810050:
### Summary 

Template page throws console warning on empty templates

<img width="2529" height="1315" alt="Image" src="https://github.com/user-attachments/assets/a13f3691-a7da-493c-8a2e-b80961607fad" />

**Root cause**:

what is happening - when we open templates page if there is no templates props.template is passed as {} by default 
That is why props.template.body is undefined 
- Vue still wraps it as Porxy(Object) {}
- the computed runs --> console.warn("template body is undefined") throws every time

```bash
const mountTest = computed(() => {
  if (!props.template?.body) {
    console.warn('Template body is undefined:', props.template);
    return {};
  }

  const test = { ...props.template.body };
  if (!test.testType && props.template.body.testType) {
    test.testType = props.template.body.testType;
  }
  return test;
});
```

**Another worse thing is this** :
this watcher runs immediately:

```bash
watch(
  () => props.template,
  (newTemplate) => {
    isMyTemplate.value =
      newTemplate?.header?.templateAuthor?.userDocId === user.value?.id;
    localTest.value = mountTest.value ? { ...mountTest.value } : null;
  },
  { immediate: true, deep: true }
);
```

that is why we are seeing it every navigation/refresh
- On mount 
- with empty template
- mountTest runs
- warning logged

PR #1402, opened 2026-01-22
fix: console warning when template body is undefined

rahul810050:
### pr purpose

this pr fixes console warning that happens when navigating to templates page with no templates available

**Fixed issue**: #1401

### Fix added:

- added proper guards for empty template state and prevents mountTest and the watcher from running when no template body is present

> this removes unnecessary console warnings and correctly handles the empty templates state

**Before**
image

**After**
image


rahul810050:
@marcgc21 @KarinePistili Please review it whenever you get a chance

marcgc21 (approved):

PR #1445, opened 2026-01-25
feat: Enhance `pr-auto-label` .

DivyanshuVortex:
## Description
Fixes an issue where the PR auto-labeler would assign incorrect labels or fail to remove them, especially in large PRs or after merges from `develop`.
--------------------------------
Screenshot (480)

--------------------------------
## Changes
-   **Fix Pagination**
-   **Robust Label Removal**: Added logic to explicitly **remove** component labels (e.g., `testing`, `ci/cd`, `backend`) if the corresponding files are no longer present. Previously, these labels would stick even after files were reverted.
-   **Filter Noise**: `package-lock.json` and `yarn.lock` are now ignored when calculating PR size and complexity, providing a more accurate "Size" label.

### DIFF
--- a/.github/workflows/pr-auto-label.yml
+++ b/.github/workflows/pr-auto-label.yml
@@ -32,28 +32,32 @@ jobs:
             const prBody = pr.body || "";
             const labelsToAdd = [];
 
-            // Get files changed in PR
-            const { data: files } = await github.rest.pulls.listFiles({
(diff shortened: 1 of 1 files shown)


DivyanshuVortex:
During the merge of develop in branch[ Ref - img ]... at the time of removing only two label was removed and the testing , cicd left as it is. ....

PR #1470, opened 2026-01-26
Dashboard: real data + new user onboarding

sagar-h007:
This PR adds real user data to the dashboard and improves onboarding for new users. It removes hardcoded/stub data, links studies to users, and shows only the user’s actual studies and sessions. It also adds a simple onboarding prompt when no studies exist to help first-time users get started.

Overall, this makes the dashboard more accurate, useful, and less fake.

### After:

https://github.com/user-attachments/assets/629345ad-9123-41b4-a96b-8e81381867f3

## FUTURE CHANGES:
We can remove create study in sidebar and add: https://github.com/ruxailab/RUXAILAB/pull/1461

### DIFF
--- a/src/app/plugins/locales/en.json
+++ b/src/app/plugins/locales/en.json
@@ -1012,9 +1012,9 @@
       "progress": "Progress",
       "updated": "Last Updated on",
       "version": "Version:",
-      "noTests": "No tests found",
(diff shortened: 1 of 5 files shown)


marcgc21:
Not interested on this kind of features right now.

PR #1504, opened 2026-01-28
fix(profile): handle optional fields

gopikrishna-20:
fixes #1490 

- Prevented crashes caused by optional/null profile fields
- Hardened validation and reactive access
- Ensured profile view renders safely with incomplete user data


marcgc21:
Looks like too much GPT

PR #1506, opened 2026-01-28
Fix: Active Studies not updating after study creation

rakshityadav1868:
## Problem
When users created a new study, it was not appearing in their "Active Studies" list on the dashboard in real-time. The studies would only appear after a manual page refresh, creating a poor user experience and confusion about whether the 
study was actually created.

## Root Cause
The study creation process had two main issues:
1. After creating a study in the database, it wasn't being linked to the user's `myTests` collection

### 1. **StudyController.js** - Enhanced `updateStudy()` method
- Added logic to link studies to users when they're created
- When `testAdmin` and `userDocId` are present, automatically adds the study to 
  the user's `myTests` collection with a `createdAt` timestamp
- Keeps the study update logic centralized in one place

### 2. **Study.js Store** - Enhanced `createStudy()` action
- After creating a study, calls `updateStudy()` to link it to the user
- Commits the updated studies list to the Vuex store so the UI updates in 
  real-time without requiring a page refresh

## Files Changed
- `src/controllers/StudyController.js` - Enhanced `updateStudy()` with user linking
- `src/store/modules/Study.js` - Enhanced `createStudy()` action with refresh logic

## BEFORE:
https://github.com/user-attachments/assets/59e2aae0-fb5d-46db-9df9-cf2e3994a580

## AFTER: 
https://github.com/user-attachments/assets/9e4d14e4-d0ac-4ad4-8002-3c78a75fbd2e

### DIFF
--- a/src/controllers/StudyController.js
+++ b/src/controllers/StudyController.js
@@ -78,7 +78,7 @@ export default class StudyController extends Controller {
       await super.update(COLLECTION, payload.id, payload.toFirestore())
       
       // Link study to user if testAdmin is present (for new creation)
-      if (payload.testAdmin && payload.testAdmin.userDocId) {
+      if (payload.testAdmin?.userDocId) {
         const userDoc = await super.readOne('users', payload.testAdmin.userDocId)
         const userData = userDoc.data()
         userData.myTests = userData.myTests || {}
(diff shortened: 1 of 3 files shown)


marcgc21:
It looks that is already working on develop. Closing it!

PR #1520, opened 2026-01-29
FIX : Media file sizes are not persisting, causing inaccurate storage reporting

### LINKED ISSUE #1519: Bug: Media file sizes are not persisting, causing inaccurate storage reporting
`Storage` section in the admin dashboard displays incorrect or fallback data instead of the actual file size, making it impossible to accurately track storage usage for user tests

Image

- Expected Behavior
When a recording is stopped, the file size (in bytes) should be captured.
This size should be persisted to the database alongside the media URL.
The Admin Dashboard > Storage section should display the exact size of the uploaded file.

- Actual Behavior
Media URLs are saved, but the associated size data is lost or set to null during the save process.
The UI falls back to hardcoded estimates or shows inaccurate data.

DivyanshuVortex:
FIX : #1519 

### Title
Fix: Persist Media Sizes for Accurate Storage Reporting

------------
After -

https://github.com/user-attachments/assets/fddb3d09-eeaa-4fc4-8e38-7614ea5795fc

Before -

https://github.com/user-attachments/assets/acf2bbe6-971d-454f-994c-df950b838e00

### Description
This PR fixes an issue where media file sizes were not consistently stored in Firestore. As a result, storage usage calculations were incorrect and some media entries contained null size values.

The update ensures media sizes are captured at recording time, propagated through state correctly, and stored in explicit, Firestore-safe fields.


DivyanshuVortex:
@marcgc21 sir , please review my pr

jvJUCA (approved):
LGTM, is clear de diference when we see the real size now, thanks for your contribution!

PR #1533, opened 2026-01-30
Make Firebase emulator ports configurable via environment variables

### LINKED ISSUE #1511: Hardcoded Firebase emulator ports in frontend make local setup fragile
While setting up the project locally, I noticed that Firebase emulator ports (Auth, Firestore, Functions, Storage) are currently hardcoded in the frontend Firebase initialization code. 
[/src/app/plugins/firebase/index.js](https://github.com/ruxailab/RUXAILAB/blob/develop/src/app/plugins/firebase/index.js)

This works in standard setups, but it can cause friction on systems where common dev ports (e.g. 8080 / 8081) are already in use (for example due to Docker’s proxy or other local services). In those cases, contributors need to modify source code just to run the project locally.

Would it make sense to make the frontend emulator ports configurable via environment variables (e.g. `VUE_APP_FIRESTORE_EMULATOR_PORT`,` VUE_APP_AUTH_EMULATOR_PORT`, etc.) instead of hardcoding them?

This could:

- improve local developer experience for contributors

- keep firebase.json as the source of truth for emulator ports, while allowing frontend flexibility

- This would only affect local development, since emulator wiring is already guarded by `VUE_APP_USE_EMULATORS`

Happy to submit a PR if this approach aligns with the project direction.

sohampirale:
Fixes #1511

### Summary

This PR removes hardcoded Firebase emulator ports from the frontend and makes them configurable via environment variables.

### Changes

- Load Auth, Firestore, Functions, and Storage emulator ports from `process.env`
- Provide sensible defaults that match the current `firebase.json` configuration
- Document available emulator env vars in `.env.example`
- Align frontend Functions emulator port with `firebase.json`

### Why

Hardcoded ports make local setup fragile on systems where common dev ports are already in use (e.g. Docker, other local services).  
This change improves local developer experience without affecting production behavior.

### Notes

- Emulator wiring remains gated behind `VUE_APP_USE_EMULATORS`


sohampirale:
Hello @marcgc21 can you review this?

KarinePistili (approved):

PR #1536, opened 2026-01-30
feat(storage): add helper to backfill storage usage for older studies

### LINKED ISSUE #1522: feat(storage): create a helper function to update storage from older studies

SuyashJain17:
## Summary
This PR adds a backend helper to calculate and backfill `storageUsageMB` for users with older studies where this field was not previously populated. This ensures storage quota tracking and dashboard values remain accurate and consistent.

Fixes #1522

## What’s changed
- Added a shared storage calculation utility that mirrors the estimation logic used in the frontend (`StorageInfo.vue`)
- Introduced a new `updateStorageUsage(userId)` method in `UserController`
- The controller:
  - Fetches the user and their tests
  - Explicitly resolves answer documents (since tests only store `answersDocId`)
  - Calculates total storage usage using estimated media sizes
  - Updates `storageUsageMB` in Firestore
- Added unit tests covering storage calculation edge cases

## Why this approach
- Keeps frontend and backend storage calculations consistent
- Avoids expensive file metadata lookups during bulk backfill
- Safely recalculates storage usage without modifying existing study data

## Testing
- Added unit tests for the storage calculation utility
- Verified correct handling of empty users, nested answers, and missing properties

## Follow-ups
The new helper can be invoked via a migration script, cloud function, or admin action to backfill storage usage for existing users.

### DIFF
--- a/src/features/auth/controllers/UserController.js
+++ b/src/features/auth/controllers/UserController.js
@@ -156,10 +156,7 @@ export default class UserController extends Controller {
       userToUpdate.notifications[notificationIndex].readAt = Date.now()
 
       // Save updated user data to Firestore
-      await this.update(
-        userToUpdate.id,
-        userToUpdate.toFirestore(),
-      )
+      await this.update(userToUpdate.id, userToUpdate.toFirestore())
       return userToUpdate
     } else {
       // Notification was not found in the array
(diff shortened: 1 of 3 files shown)


marcgc21:
The helper function is only used as a test and it needs to be used to recalculate the size. Closing it because it needs further improvements.

PR #1540, opened 2026-01-31
fix: show End Call button for observator in moderated test

### LINKED ISSUE #1527: fix(user-test): on moderated test, observator can not see "End call" button

sshivanshg:
Fixes #1527

Changes:
- Show single "End Call" button for both moderator and observator when call is active
- Remove separate "Leave Call" button for participants
- Update endCall() to clear room for everyone regardless of who ends the call


marcgc21:

Issue #1526, opened 2026-01-30
fix(user-test): on moderated test, if a user is disconnected, he/she can't join the room again


marcgc21:

rakshityadav1868:
working on this !!

sohampirale:
checking this out and starting to work on it.

Santosh69:
working on this.

sshivanshg:
hey @marcgc21 check this out and tell me if there are any issues?

PR #1541, opened 2026-01-31
fix(user-test): allow disconnected user to rejoin moderated call

sshivanshg:
- Track room open state (roomIsOpen) so UI can show rejoin when room is still active
- Set roomReady = false in leaveRoom() for non-moderator so they can re-join
- Clear peers map in leaveRoom() so re-join creates fresh connections
- Add Rejoin call / Rejoin to observe buttons for participant and observator
- rejoinRoom() cleans stale state then calls joinRoom() again
- Hide 'Waiting for moderator' when room is open; show rejoin UI when left

fixes - #1526

### DIFF
--- a/src/ux/UserTest/components/VideoCall.vue
+++ b/src/ux/UserTest/components/VideoCall.vue
@@ -121,9 +121,9 @@
         </div>
       </v-col>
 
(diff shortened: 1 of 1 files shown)


marcgc21:

PR #1553, opened 2026-02-01
fix(user-test): sync camera/mic/screen state to other participants in…

### LINKED ISSUE #1528: fix(user-test): on moderated test, if moderator or user stops sharing screen or microphone do not display information for the other participants
When the user stops sharing camera or microphone it appears on your own UI, but not for the others UI. So if someone presses that buttons, need to indicate to the other participants

sshivanshg:
… moderated call

Fixes #1528

- Store hasCamera, hasMicrophone, hasScreenShare in Firebase participant node on join
- Update participant node when user toggles camera, microphone, or screen share
- Add getPeerHasCamera/getPeerHasMicrophone helpers reading from participants
- Show camera-off overlay and mic-muted indicator on remote video tiles
- participantsList uses remote hasCamera/hasMicrophone/hasScreenShare from Firebase

### DIFF
--- a/src/ux/UserTest/components/VideoCall.vue
+++ b/src/ux/UserTest/components/VideoCall.vue
@@ -69,6 +69,23 @@
                 playsinline
                 class="video-element"
               ></video>
+              <!-- Camera off overlay (synced from Firebase) -->
+              <div
+                v-if="!getPeerHasCamera(userId)"
+                class="camera-disabled-overlay"
+              >
+                <v-icon size="64" color="white" class="mb-2"
+                  >mdi-video-off</v-icon
+                >
+                <p class="text-white">Camera is off</p>
+              </div>
+              <!-- Mic muted indicator (synced from Firebase) -->
+              <div
+                v-if="!getPeerHasMicrophone(userId)"
+                class="mic-muted-indicator"
+              >
+                <v-icon size="24" color="white">mdi-microphone-off</v-icon>
+              </div>
               <div class="video-label">{{ getPeerName(userId) }}</div>
             </div>
           </div>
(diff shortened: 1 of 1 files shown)


src/ux/UserTest/components/VideoCall.vue
```diff
@@ -916,14 +934,33 @@ const participantsList = computed(() => {
...
+          : (p.hasMicrophone !== undefined ? p.hasMicrophone : true),
+      hasScreenShare: !!(p.hasScreenShare === true),
     })
   })
 
   return list
 })
 
+// Helpers for remote peer media state (synced via Firebase)
+const getPeerHasCamera = (userId) => {
```
Swastik19Nit: I see the helper functions use a fallback:
const getPeerHasCamera = (userId) => {
  const p = participants.value[userId]
  if (!p) return true
  return p.hasCamera !== undefined ? p.hasCamera : true
}

Can you tell me why the default is 'true'. If a participant's data has not loaded yet or firebase is slow the UI will  assume they have a camera on or potentially showing a black box before the "Camera is off" overlay appears. It might be better to default to false or a "loading" state to prevent UI flickering. Justify your approach and tell me.

Swastik19Nit, line 953: I see the helper functions use a fallback.

```js
const getPeerHasCamera = (userId) => {
  const p = participants.value[userId]
  if (!p) return true
  return p.hasCamera !== undefined ? p.hasCamera : true
}
```
Can you tell me why the default is `true` . If a participant's data has not loaded yet or Firebase is slow, the UI will assume they have a camera on r potentially showing a black box before the "Camera is off" overlay appears. It might be better to default to `false` or a loading state to prevent UI flickering. justify your approach and tell me.

marcgc21:

PR #1558, opened 2026-02-01
fix : isStartTestDisabled returned early for admins without actually updating the reactive variable to false

DivyanshuVortex:
- The `watchEffect` responsible for controlling `isStartTestDisabled` returned early for **admin users** without explicitly updating the reactive state to `false` when all validation checks passed.

- Because `watchEffect` **ignores return values**, the logic that intended to allow admins to proceed (`return false`) had no effect on the actual UI state.

---
After - 
Screenshot (511)

Before - 
Screenshot (512)


DivyanshuVortex:
hey @jvJUCA and @marcgc21 sir , please review this PR , this is auto-close as merge the develop branch into this branch

PR #1566, opened 2026-02-02
fix: remove video event listeners in SessionAnalyticsDialog on unmount

### LINKED ISSUE #1565: [🐞 BUG]: Video event listeners not removed in SessionAnalyticsDialog.vue
### Description 📝

In `SessionAnalyticsDialog.vue`, three native DOM event listeners (`loadedmetadata`, `play`, `pause`) are added to the screen-recording video element (`mainVideo2`) in `onMounted`, but they are never removed in `onBeforeUnmount`. Only `cancelAnimationFrame(rafId)` is called on unmount. This violates the usual "clean up what you add" pattern and can lead to listener leaks or stale closures if the component stays mounted while the dialog is closed, or if the dialog implementation changes. The handlers are anonymous, so they cannot be removed without storing references.

### Link 🔗

N/A (Issue observed during code inspection and local testing)

### Steps to Reproduce 🔄

1. Go to a User Test (moderated or unmoderated) that has at least one task with screen recording.
2. Open the session / task analytics view where you can open "Task Analysis" for a task.
3. Click to open the Task Analysis dialog (SessionAnalyticsDialog) so the screen video and timeline are shown.
4. Play/pause the screen video (optional => confirms listeners are attached).
5. Close the dialog (and optionally navigate away so the component unmounts).
6. Inspect the code: in `onBeforeUnmount` only `cancelAnimationFrame(rafId)` runs; there is no `removeEventListener` for the three video listeners added in `onMounted`.

### Screenshots 📸

N/A — this is a code-level bug (missing cleanup), not a visible UI bug. The relevant code is in `SessionAnalyticsDialog.vue` lines 227–245: `onMounted` adds three `addEventListener` calls; `onBeforeUnmount` only calls `cancelAnimationFrame(rafId)`.

### Expected Behavior 🤔

When the component unmounts, all native event listeners added in `onMounted` should be removed in `onBeforeUnmount`. Specifically: the same function references used in `addEventListener('loadedmetadata', ...)`, `addEventListener('play', ...)`, and `addEventListener('pause', ...)` should be passed to `video.removeEventListener(...)` before the component is destroyed, in addition to calling `cancelAnimationFrame(rafId)`.

### Actual Behavior 😱

Only `cancelAnimationFrame(rafId)` is called in `onBeforeUnmount`. The three `addEventListener` calls in `onMounted` (lines 231–243) are never paired with `removeEventListener`. The handlers are anonymous, so they cannot be removed without refactoring to use stored function references.

### Environment 🌍

- Vue component: `src/ux/UserTest/components/dialogs/SessionAnalyticsDialog.vue`
- Git branch: develop 
- OS: Any (bug is in component code, not environment-specific)
- Browser: Any (bug is in component code, not environment-specific)
- Vue: 3.x, Composition API (script setup)

### Additional Information ℹ️

**Suggested fix:** Store named handler functions (e.g. `onLoadedMetadata`, `onPlay`, `onPause`) and in `onBeforeUnmount` call `video.removeEventListener('loadedmetadata', onLoadedMetadata)` (and same for `play` / `pause`) if `mainVideo2.value` is still set, then call `cancelAnimationFrame(rafId)`. This ensures proper cleanup and avoids listener leaks or stale closures.

harsh04044:
# Fixes #1565 

## Description

The Task Analysis dialog was leaving native video event listeners attached to the screen video element after the dialog closed. Listeners for `loadedmetadata`, `play`, and `pause` were added in `onMounted` but never removed in `onBeforeUnmount`, which could lead to leaks and unexpected behavior when the dialog is opened and closed repeatedly.

## What changed

- **SessionAnalyticsDialog.vue**  
  - Reused the existing `onMetadataLoaded` handler for the `loadedmetadata` listener so it can be removed on unmount.  
  - Added named handlers `onVideoPlay` and `onVideoPause` (same behavior as before) and used them for the `play` and `pause` listeners.  
  - In `onBeforeUnmount`, the component now removes all three listeners from the video element (when it exists) using the same handler references, then calls `cancelAnimationFrame(rafId)` as before.

No template changes, no changes to other components or files. Behavior (timeline, play/pause, seeking) is unchanged.

## How to verify

1. **Behavior unchanged** — Open the Task Analysis dialog, play and pause the screen video, and use the timeline. Timeline and play state should behave as before.
2. **Cleanup on close** — Close the dialog (component unmounts). The three listeners are now removed in `onBeforeUnmount`, so no listeners are left attached. Optionally, in devtools you can inspect the video element before closing to confirm only the expected listeners, or open/close the dialog multiple times and confirm no errors or odd behavior.

### DIFF
--- a/src/ux/UserTest/components/dialogs/SessionAnalyticsDialog.vue
+++ b/src/ux/UserTest/components/dialogs/SessionAnalyticsDialog.vue
@@ -198,6 +198,16 @@ function onTimeUpdate(event) {
   videoCurrentTime.value = video.currentTime
 }
 
+function onVideoPlay() {
+  isPlaying.value = true
+  updateLoop()
+}
+
+function onVideoPause() {
+  isPlaying.value = false
+  cancelAnimationFrame(rafId)
+}
+
 const togglePlay = () => {
   const video = mainVideo2.value
   if (!video) return
@@ -228,21 +238,19 @@ onMounted(() => {
   const video = mainVideo2.value
   if (!video) return
 
-  video.addEventListener('loadedmetadata', () => {
-    videoDuration.value = video.duration
-  })
-
-  video.addEventListener('play', () => {
-    isPlaying.value = true
-    updateLoop()
-  })
-
-  video.addEventListener('pause', () => {
-    isPlaying.value = false
-    cancelAnimationFrame(rafId)
-  })
+  video.addEventListener('loadedmetadata', onMetadataLoaded)
+  video.addEventListener('play', onVideoPlay)
+  video.addEventListener('pause', onVideoPause)
 })
-onBeforeUnmount(() => cancelAnimationFrame(rafId))
+onBeforeUnmount(() => {
+  const video = mainVideo2.value
+  if (video) {
+    video.removeEventListener('loadedmetadata', onMetadataLoaded)
+    video.removeEventListener('play', onVideoPlay)
+    video.removeEventListener('pause', onVideoPause)
+  }
+  cancelAnimationFrame(rafId)
+})
 </script>
 
 <style scoped>


harsh04044:
@marcgc21  can you please review this?

jvJUCA (approved):
Maybe worth centralizing the add/remove logic in a helper or composable to avoid repetition if this pattern shows up elsewhere. Also, cancelAnimationFrame could live in the same cleanup block for consistency. Other than that, cleanup looks solid.

Thanks for contributing!

PR #1567, opened 2026-02-02
Fix :  Progress Calculation Refactor ( Moderated_test )

DivyanshuVortex:
### Summary
This PR refactors the progress calculation logic in `ModeratedTestView` to make test progress tracking more reliable, predictable, and easier to reason about across moderated test flows, while **aligning the progress semantics with the existing unmoderated test progress logic** to ensure consistent behavior and reporting across both test types.

---

### Visual comparison

- **Before**  
Before progress logic

- **After**  
Screenshot (507)

### Why this change
The previous implementation mixed task-level and phase-level progress, which led to inconsistent and sometimes misleading progress percentages, especially when task counts or configurations changed.


src/ux/UserTest/views/ModeratedTestView.vue
```diff
@@ -1156,23 +1156,37 @@ const validate = (object) => {
...
+    if (!localTestAnswer) return 0
+    const totalSteps = 4
     let completedSteps = 0
-    if (localTestAnswer.consentCompleted) completedSteps += 1
-    if (localTestAnswer.preTestCompleted) completedSteps += 1
-    if (localTestAnswer.postTestCompleted) completedSteps += 1
 
-    if (Array.isArray(localTestAnswer.tasks)) {
-      completedSteps += localTestAnswer.tasks.filter((t) => t.completed).length
+    if (localTestAnswer.preTestCompleted) completedSteps++
```
Swastik19Nit: I can suggest a more Vue-idiomatic way to write that same logic, as it removes the `for` loop in favour of `.every()` or `.filter()`, which I think is generally preferred in this codebase for readability.

```js
const calculateProgress = () => {
  try {
    if (!localTestAnswer) return 0;
    const TOTAL_CATEGORIES = 4; 
    let completedCategories = 0;

    if (localTestAnswer.consentCompleted) completedCategories++;
    if (localTestAnswer.preTestCompleted) completedCategories++;

    // you can now check if all tasks are finished
    const tasks = localTestAnswer.tasks;
    if (Array.isArray(tasks) && tasks.length > 0) {
      const allDone = tasks.every(t => t?.completed || t?.attempted);
      if (allDone) completedCategories++;
    }

    if (localTestAnswer.postTestCompleted) completedCategories++;

    const progress = (completedCategories / TOTAL_CATEGORIES) * 100;
    localTestAnswer.progress = progress;
    return progress;
  } catch (error) {
    console.error('Progress calculation failed:', error);
    return 0;
  }
};
```

src/ux/UserTest/views/ModeratedTestView.vue
```diff
@@ -1156,23 +1156,37 @@ const validate = (object) => {
 
 const calculateProgress = () => {
   try {
-    const totalSteps = test.value?.testStructure?.userTasks?.length || 0
-    if (totalSteps === 0) return 0
-
+    if (!localTestAnswer) return 0
+    const totalSteps = 4
```
Swastik19Nit: you had const totalSteps = 4 so ensure that this variable is actually defined within the scope of this function or as a constant at the top of the script. If `totalSteps` is undefined when this runs, `progressPercentage` can return NaN

DivyanshuVortex: @Swastik19Nit Thanks for the feedback. I’ve followed the same structure that’s used in the unmoderated test, which is why I implemented the logic this way. I understand and agree with your point of view, though. If the maintainers prefer a different structure, I’m happy to refactor it accordingly. Thanks again.

KarinePistili (approved):
Tested and it worked well. Thank you for the contribution.

PR #1573, opened 2026-02-02
Fix: handle missing pre-test and post-test sections in task completion logic

rakshityadav1868:
## Overview
Fix a critical bug in the moderated test where completing tasks fails with `TypeError: Cannot read properties of undefined (reading 'value')` when the test structure doesn't include pre-test or post-test sections.

## Problem Statement
When a test is created without pre-test or post-test sections, the `items.value` array structure changes:
- **With pre-test**: `items.value[0]` = Pre-test, `items.value[1]` = Tasks
- **Without pre-test**: `items.value[0]` = Tasks (index mismatch!)

The `completeStep()` function was hardcoded to access `items.value[1]` for tasks, causing undefined reference errors when pre-test or post-test are missing.

## Solution
Added safe optional chaining (`?.`) checks before accessing nested `items.value` properties:

## BEFORE:

https://github.com/user-attachments/assets/7109c458-c181-412d-a180-f3ed729bb976

## AFTER:

https://github.com/user-attachments/assets/52ce8f06-e197-4559-bf02-75b1bbae785a

### DIFF
--- a/src/ux/UserTest/views/ModeratedTestView.vue
+++ b/src/ux/UserTest/views/ModeratedTestView.vue
@@ -994,16 +994,18 @@ const completeStep = async (id, type, userCompleted = true) => {
         return
       }
       localTestAnswer.tasks[id].completed = userCompleted
-      items.value[1].value[id].icon = 'mdi-check-circle-outline'
+      if (items.value[1]?.value?.[id]) {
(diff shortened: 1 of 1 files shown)


jvJUCA (approved):
Your pull request has been accepted! Thank you for your valuable contribution.

PR #1612, opened 2026-02-05
fix(router): require token for accessibility preview (Issue 7)

### LINKED ISSUE #1606: [🐞 BUG]: Accessibility preview route bypasses authentication
### Description 📝

In src/app/router/index.js (lines 29–35), the router guard allows any path that contains both /accessibility/ and /preview/ to proceed without any authentication or token check. The guard returns next() immediately, before checking user or meta.authorize. The accessibility manual preview route is /accessibility/manual/preview/:id/:userId?/:token? and loads assessment data by test ID and optionally user ID. The route meta comment says "Allow public access with token," but the guard does not validate the token or require it , so unauthenticated users can open preview URLs (e.g. /accessibility/manual/preview/{testId}) and view assessment data if they know or guess the test ID. This bypass contradicts the stated intent and can expose accessibility assessment data to unauthorized access when IDs are guessable or leaked.

### Link 🔗

N/A (Issue observed during code inspection and local testing)

### Steps to Reproduce 🔄

1. Ensure you are not signed in (or use an incognito window with no session).
2. Open a URL that contains /accessibility/ and /preview/, e.g. /accessibility/manual/preview/{someTestId} (replace {someTestId} with a valid test ID if you have one from another session).
3. Observe that the router allows access and the preview page loads without prompting for sign-in or validating a token.
4. Inspect src/app/router/index.js: the guard checks to.path.includes('/accessibility/') && to.path.includes('/preview/') and returns next() with no auth or token check.

### Screenshots 📸

N/A — this is a router/security behavior; the relevant code is in src/app/router/index.js lines 29–35. 

### Expected Behavior 🤔

If the intent is "public access with token," the guard should allow access to accessibility preview routes only when a valid token is present (e.g. in route params or query) and optionally tie that token to the test (and user). If the intent is that preview should never be public, the guard should not bypass auth for these paths; users should be required to sign in and have the appropriate role/test ownership before viewing the preview. In both cases, the guard should not allow unauthenticated access to preview URLs without a validated token or auth.

### Actual Behavior 😱

Any path containing /accessibility/ and /preview/ is allowed through the router guard with no authentication check and no token validation. Unauthenticated users can open URLs like /accessibility/manual/preview/{testId} and, if the test ID is valid, the preview page loads and can display assessment data. The route meta says "Allow public access with token" but the guard does not enforce the token, so the current behavior is "allow public access without token."

### Environment 🌍

- Vue component / file: src/app/router/index.js
- Git branch: develop
- OS: any
- Browser: any
- Vue: 3.x, Vue Router 4

### Additional Information ℹ️

Suggested fix: (1) If preview must be public only with a shareable link, require a token in the path or query (e.g. to.params.token or to.query.token), validate it (e.g. against Firestore or a backend), and allow access only when valid. Optionally restrict Firestore/Storage rules so assessment data for a test is readable only when the request includes or is tied to a valid token. (2) If preview is not meant to be public, remove the accessibility-preview bypass in the guard and require authentication (and optionally role or test ownership) for the preview route so it behaves like other protected routes.

harsh04044:
## Problem

Accessibility preview routes had no token enforcement:

- Any path containing `/accessibility/` and `/preview/` bypassed authentication
- Unauthenticated users could open preview URLs (e.g. `/accessibility/manual/preview/{testId}`) and view assessment data with only the test ID
- Route meta stated "Allow public access with token," but the guard did not check for a token

## Solution

Updated the router guard so accessibility preview access is gated on a token:

- Token can be in path params (`/preview/:id/:userId?/:token?`) or query (`?token=...`)
- If token is present → allow access (shareable links keep working)
- If token is absent on a preview path → redirect via existing logic (sign-in for guests, dashboard for logged-in users)
- No change to other routes; existing auth and authorize behavior unchanged

## Changes

**Added**

- `accessibilityPreviewGuard.js` – pure helper that decides allow/redirect for preview paths (used by the guard and easy to test later)

**Updated**

- `src/app/router/index.js` – guard now calls the helper and allows only when token is present; otherwise redirects

## Benefits

- Shareable preview links with token still work for unauthenticated users
- Preview URLs without a token no longer expose assessment data
- Single, clear place for preview access logic
- No changes to route definitions, preview component, or Firestore

## Testing

- Manual verification: with token → preview loads; without token → redirect to sign-in/dashboard; other routes unchanged

Fixes: #1606


harsh04044:
Implemented the token check as described. Should I add unit tests for this (allow with token, redirect without) in a follow-up commit?

harsh04044:
Note: Sorry for the small mistake in the PR title. The text “(Issue 7)” was only for my personal reference and is not related to this pull request. I added it by mistake and I’m currently unable to edit the title.

marcgc21:
From my point of view this issue could be solved more easily... let me clarify something, we don't have distinctions about if some study is from heuristics, accessibility, or wathever...

So I think that we should only remove those weird if's that allows public access. Can you fix that? @harsh04044

harsh04044:
@marcgc21 why this pr is closed?

PR #1638, opened 2026-02-06
fix(lint):fix linting for `src/ux/CardSorting`

### LINKED ISSUE #1630: style(lint): fix lint for src/shared/views folder
After run:

npx eslint src/ux/CardSorting

We should reduce the number of warnings to 0, right now it returns around 29 warnings, most of them related to lint or i18n problems.

DivyanshuVortex:
Fixes : #1630

----------------------------
Before ->
Screenshot (514)

After -> 

Screenshot (515)


DivyanshuVortex:
@marcgc21 sir,

The code `duplication` is around 9%. Reducing it would require significant refactoring,but this PR is only for lint fix . I’d appreciate your thoughts on whether I should address it here or handle it separately *

PR #1648, opened 2026-02-08
Fix : Enhanced Moderated Testing Experience & adds a Cleanup function. 

### LINKED ISSUE #1526: fix(user-test): on moderated test, if a user is disconnected, he/she can't join the room again

DivyanshuVortex:
fix : #1526 On moderated test, if a user is disconnected, he/she can't join the room again

The main issue if the moderated refresh then all the other disconnect from the room as the room is ONLY dependent on the moderator .

### Before this
- Room was dependent on the moderator. If the moderator refreshed, all connected users were disconnected and the room was deleted.
- Media preferences were not saved on refresh.
- Risk of ghost data in RTDB.
- No heartbeat logic.

```  
 Moderator connects 
      ↓
Room created in RTDB
      ↓
Participants join
      ↓
Moderator refreshes
      ↓
Room deleted from RTDB
      ↓
All participants disconnected
  ```

------------
## NEED TO DEPLOY THIS CLEANUP FUNCTION
--------------

### After this
- Room is no longer tied to a specific user. It is removed from RTDB only when:
  - The moderator explicitly clicks **End Call**, or
  - The scheduled cleanup Cloud Function runs.
- Refresh and re-join are supported for all participants.
- Media preferences persist across refresh. [ change in the RTDB schema a bit ]
- Heartbeat logic ensures active session tracking.
- Scheduled cleanup function removes RTDB sessions not updated for more than 6 hours.

```
Moderator connects
      ↓
Session created in RTDB
      ↓
Participants join
      ↓
Any user refreshes
      ↓
Session remains in RTDB
      ↓
User reconnects using same session
      ↓
RTDB clear by the backend function automatically 

```
-----------------------

```
            +----------------------+
            |   Session Created    |
            +----------+-----------+
                       |
                       v
            +----------------------+
            |   Heartbeat Updates  |
            |  (active users)      |
            +----------+-----------+
                       |
         +-------------+-------------+
         |                           |
         v                           v
+------------------+       +----------------------+
| End Call Clicked |       | No update > 6 hours  |
+--------+---------+       +----------+-----------+
         |                            |
         v                            v
   Session Deleted             Cleanup Function
                                Deletes Session

```
-----------------------------------

Screenshot (518)

Video link - https://drive.google.com/file/d/1O_91RbqUOHabYL98CLraGENdb0dAxNpg/view?usp=sharing

This approach keeps the session alive even if someone refreshes or disconnects, so the system doesn’t break or delete everything by mistake.

A Cloud Function runs every 6 hours to clean up old or unused data from the Realtime Database. This prevents unused session data from piling up and keeps the database clean and organized.

Because the session data is preserved properly, users can refresh or rejoin the call without losing their settings, such as media preferences. Overall, this makes the system more stable, easier to maintain, and less prone to errors caused by stale or leftover data.


DivyanshuVortex:
fixes : #1525

DivyanshuVortex:
fixes : #1527  - on moderated test, observator can not see "End call" button , now the observer only able to see leave call ..


https://github.com/user-attachments/assets/179c4d63-a9b6-4f85-94a7-664b2718cc1f

DivyanshuVortex:
When the moderator is disconnect from a call for more than 5 mins , a alert box is showing to all theh participant on the call . 

Screenshot (524)

PR #1676, opened 2026-02-12
Fix: Admin redirected to test steps on refresh and after step completion in moderated sessions

rakshityadav1868:
Admin was being redirected to test steps during refresh. Fixed Firebase listener to keep admin always in video call during active session.

**Changes:**
- Added condition in Firebase listener to ensure `displayVideoCallComponent = true` for admin throughout session
- Admin never sees test step components (ConsentStep, PreTestStep, TaskStep, etc.)

**File Modified:**
- `src/ux/UserTest/views/ModeratedTestView.vue` 

## BEFORE:

https://github.com/user-attachments/assets/356be7fb-8aee-4b41-9cfd-f948d973706e

## AFTER:

https://github.com/user-attachments/assets/637f2632-7c50-4f2c-ac95-023e934b6c61

### DIFF
--- a/src/ux/UserTest/views/ModeratedTestView.vue
+++ b/src/ux/UserTest/views/ModeratedTestView.vue
@@ -643,7 +643,8 @@ const { t } = useI18n()
 // Data variables
 
 onBeforeUnmount(() => {
-  if (moderatorDisconnectTimeout.value) clearTimeout(moderatorDisconnectTimeout.value)
+  if (moderatorDisconnectTimeout.value)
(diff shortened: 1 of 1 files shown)


DivyanshuVortex:
@rakshityadav1868 
Thanks for catching this.

It was fixed earlier, but I missed it during the heartbeat refactor — that’s on me. I’ve tested the changes locally and it’s working well. Appreciate you pointing it out.

KarinePistili (approved):

PR #1705, opened 2026-02-19
fix: resolve manager dashboard metrics reading from non-existent data source

### LINKED ISSUE #1704: [🐞 BUG]:  Manager dashboard Beta components always show 0 for all metrics
### Description 📝

The Manager dashboard's Beta components (`StudyOverview`, `ParticipantsInfo`, `TasksInfo`, `StorageInfo`) display `0` for all metrics (Total Participants, Completed, In Progress, Average Time, Task Completion Rate, Success Rate, Avg. Task Duration, media file counts, etc.) even when participants have submitted answers.

The root cause is that all four components read answer data from `props.test.answers`, but the `Study` model (`src/shared/models/Study.js`) does not define an `answers` property. It only stores `answersDocId` — a reference to a separate document in the `answers` Firestore collection. The actual answer data is loaded into the Vuex store via the `Answer.js` module and accessed through `store.getters.testAnswerDocument.taskAnswers`.

Since `props.test.answers` is always `undefined`, the fallback `|| []` produces an empty array, causing every downstream metric to compute as `0`.

Additionally, `TasksInfo.vue` gates task iteration on `task.attempted`, which defaults to `false` in the `TaskAnswer` model. This means that even if data were flowing, metrics like completion rate and duration would be undercounted.

### Link 🔗

`/userTest/unmoderated/manager/{study_id}` — Manager view for any unmoderated study

### Steps to Reproduce 🔄

1. Go to the **Home** page and click **Create Study** to create a new Unmoderated User Test with at least one task.
2. Click on **Cooperators** and invite at least one user.
3. Have the invited cooperator (or yourself as the admin) open the test link and complete all tasks, then submit.
4. Go back to the **Manager** view for the study (`/userTest/unmoderated/manager/{study_id}`).
5. Scroll down to the Beta dashboard cards — **Study Overview**, **Participants Info**, **Tasks Info**, and **Storage Info**.
6. See that all metrics (Total Participants, Completed, In Progress, Average Time, Task Completion Rate, Success Rate, Avg. Task Duration, media counts) display `0` despite answers existing in Firestore.

### Screenshots 📸

Image
Image

### Expected Behavior 🤔

The dashboard should display accurate metrics reflecting the actual submitted test answers:
- **Total Participants** should count all users who have taken the test
- **Completed** should count users who submitted their answers
- **In Progress** should count users who started but haven't submitted
- **Average Time**, **Task Completion Rate**, **Success Rate**, and **Avg. Task Duration** should compute from actual task answer data

### Actual Behavior 😱

All metrics display `0` regardless of how many participants have submitted answers. The Beta cards appear functional but show no data because the underlying computed properties always receive an empty array.

### Environment 🌍

- Vue component: [`StudyOverview.vue`](https://github.com/ruxailab/RUXAILAB/blob/develop/src/ux/UserTest/components/manager/StudyOverview.vue), [`ParticipantsInfo.vue`](https://github.com/ruxailab/RUXAILAB/blob/develop/src/ux/UserTest/components/manager/ParticipantsInfo.vue), [`TasksInfo.vue`](https://github.com/ruxailab/RUXAILAB/blob/develop/src/ux/UserTest/components/manager/TasksInfo.vue), [`StorageInfo.vue`](https://github.com/ruxailab/RUXAILAB/blob/develop/src/ux/UserTest/components/manager/StorageInfo.vue)
- Git branch: `develop`
- OS: All OS
- Browser: All browsers

### Additional Information ℹ️

The existing `GeneralAnalytics.vue` component (under Reports/Answers) correctly reads answer data from `store.getters.visibleUserAnswers`, which is why the analytics page works while the Manager dashboard does not. The fix involves switching the four dashboard components to read from the same `Answer.js` Vuex store module.

Additionally, `ParticipantsInfo.vue` has a secondary bug in its `notStartedParticipants` calculation: it subtracts `completedAnswers` from `acceptedCooperators`, which can produce a **negative value** when the study admin takes the test without being in the cooperators list (e.g., 2 accepted cooperators − 3 completed answers = −1). The dashboard would then display "-1 Not Started".

thechillbasu:
# fix: resolve manager dashboard metrics reading from non-existent data source

## Description
The Manager dashboard's Beta components (`StudyOverview`, `ParticipantsInfo`, `TasksInfo`, `StorageInfo`) were displaying `0` for all metrics despite participants having submitted answers.

**Technical Analysis:**
The bug was traced to how all four components access answer data. Each component computed answers from `props.test.answers`:

```javascript
const answers = computed(() => {
  const testAnswers = props.test?.answers || []
  return Array.isArray(testAnswers) ? testAnswers : Object.values(testAnswers)
})
```

However, the `Study` model (`src/shared/models/Study.js`) does not define an `answers` property in its constructor. It only stores `answersDocId` — a reference ID to a separate document in the `answers` Firestore collection. The actual answer data is loaded into the Vuex store via the `Answer.js` module and accessed through `store.getters.testAnswerDocument.taskAnswers`.

Since `props.test.answers` is always `undefined`, the fallback `|| []` produces an empty array, causing every downstream metric (Total Participants, Completed, In Progress, Average Time, Completion Rate, Success Rate, Task Duration, media counts) to compute as `0`.

Additionally, `TasksInfo.vue` gated task iteration on `task.attempted`, which defaults to `false` in the `TaskAnswer` model. This meant that even if data were flowing, metrics like completion rate and average duration would be undercounted. The existing, working `GeneralAnalytics.vue` component does not check `task.attempted`.

## Solution and Reasoning
I refactored all four components to read answer data from the correct Vuex store module instead of the non-existent prop.

**Why this is the correct fix:**
- The `Answer.js` Vuex module already handles fetching and storing answer data via the `getCurrentTestAnswerDoc` action. The `GeneralAnalytics.vue` component successfully uses this pattern via `store.getters.visibleUserAnswers`.
- Rather than embedding answer data into the `Study` model (which would require changes to the data layer and violate the existing separation of concerns), switching the dashboard components to use the already-available store getter is the minimal, non-breaking fix.
- The duplicated `answerDocument + allAnswers` filter block was extracted into a new `allAnswersList` getter in `Answer.js`, so all four components now read from a single centralized source instead of each duplicating the same 10-line filter logic.
- Task iteration was aligned with `GeneralAnalytics.vue` — using `Object.values(answer.tasks || {})` without the `task.attempted` gate — ensuring metrics match the analytics page.
- `taskSuccessRate` now uses `task.completed && task.success !== false`, matching how `GeneralAnalytics.vue` computes success in `getTasksPerformance` and `calculateEfficiency`. This is forward-compatible with a potential `success` field in the `TaskAnswer` model for evaluator-verified outcomes (ISO 9241-11).
- The `notStartedParticipants` calculation in `ParticipantsInfo.vue` was also fixed: it previously subtracted `completedAnswers` from `acceptedCooperators`, which could go negative when the admin takes the test (they have an answer but aren't in the cooperators list). Now it subtracts from `totalParticipants` with a `Math.max(0, ...)` floor.

## Changes Made

| File | Change |
|------|--------|
| `Answer.js` | Added `allAnswersList` getter to centralize the duplicated answer-filtering logic from all 4 components |
| `StudyOverview.vue` | Read answers from `store.getters.allAnswersList`; align `averageCompletionTime` with `GeneralAnalytics.vue`'s `averageTimePerTask` (no `submitted` filter, no `taskTime` guard, count all tasks) |
| `ParticipantsInfo.vue` | Read answers from `store.getters.allAnswersList`; fix `totalParticipants` to use `Math.max(cooperators, answers)`; fix `notStartedParticipants` to prevent negative values |
| `TasksInfo.vue` | Read answers from `store.getters.allAnswersList`; align `overallCompletionRate` with `calculateEffectiveness`, `taskSuccessRate` with `getTasksPerformance`, and `averageTaskDuration` with `averageTimePerTask` from `GeneralAnalytics.vue` |
| `StorageInfo.vue` | Read answers from `store.getters.allAnswersList` for media file and storage calculations |

## Before and After

**Before (Bug)**
All manager dashboard metrics show `0` despite 3 participants having taken the test.
dashboard_before_1
dashboard_before_2

**After (Fix)**
Metrics correctly reflect submitted answer data — Total Participants, Completed, In Progress, Completion Rate, Success Rate, and Task Duration all display accurate values.
dashboard_after_1
dashboard_after_2

Closes #1704


thechillbasu:
@marcgc21  @jvJUCA Please have a look!

jvJUCA (approved):
LGTM, thanks for your contribution!

PR #1714, opened 2026-02-20
fix: replace hardcoded 'es' locale with useI18n, support Portuguese locale, and fix v-data-table sorting via stable keys and custom-key-sort

### LINKED ISSUE #1712: [🐞 BUG]: StorageSection dates hardcoded to 'es' locale, table sorting broken, and Portuguese dates render as hyphens
### Description 📝

The Storage page (`StorageSection.vue`) has three related bugs:

**1. Hardcoded Spanish locale for dates**
All dates in the Media Files table are formatted in Spanish (e.g., "18 de febrero de 2026") regardless of the user's selected language. The `formatDateLong` utility is called with a hardcoded `'es'` locale instead of using the i18n locale from `useI18n()`.

```javascript
// Bug: hardcoded 'es' locale
const date = formatDateLong(answer.date || test.creationDate, 'es')
```

**2. Portuguese dates render as hyphens**
When Portuguese is selected, all dates on the Storage page display as `-`. The app's i18n locale uses `pt_br`, but `Intl.DateTimeFormat` expects the standard BCP 47 `pt-BR`. The underscore format causes the API to throw, returning the fallback hyphen.

**3. Table sorting arrows change but rows don't reorder**
Clicking the column headers (Study Name, Date, Size) toggles the sort arrow direction but the rows never reorder. This is caused by two issues:

- **Duplicate `id` keys**: Multiple file types (video, audio, screen, webcam) from the same task all share the same `task.id`. The fallback `self.crypto.randomUUID()` generates a **new** ID on every computed re-evaluation, creating key instability. Vue warns about `Duplicate keys found during update`, and Vuetify cannot track rows for re-rendering.
- **String-based date values**: The `date` field stores a formatted string like `"18 February 2026"`, which sorts alphabetically, not chronologically.

### Link 🔗

`/admin?section=storage` — Storage section in the admin panel

### Steps to Reproduce 🔄

1. Go to the **Admin** panel and click **Storage** in the sidebar.
2. Change the language to **English** using the language selector in the navbar.
3. See that dates still appear in Spanish format (e.g., "18 de febrero de 2026").
4. Change the language to **Portuguese**. See that all dates turn into `-`.
5. Click on the **Study Name** column header to sort — the arrow toggles but "Test 4" does not sort before "Test 7".
6. Click on the **Date** column header — the arrow toggles but earlier dates do not move to the top.
7. Click on the **Size** column header — the arrow toggles but smaller files do not move to the top.

### Screenshots 📸

https://github.com/user-attachments/assets/aceec203-ac37-4e57-8067-8094b0021d0b

### Expected Behavior 🤔

- Dates should be formatted according to the user's selected language (e.g., "18 February 2026" for English, "18 de fevereiro de 2026" for Spanish) and render correctly for Portuguese.
- Clicking column headers should sort rows correctly:
  - **Study Name**: Alphabetical with natural number ordering (Eg: Test 4 < Test 5 < Test 7 < Test 10)
  - **Date**: Chronological (oldest first ascending, newest first descending)
  - **Size**: Numeric (smallest first ascending, largest first descending)

### Actual Behavior 😱

- Dates always display in Spanish (or `-` in Portuguese) regardless of language selection.
- Sort arrows toggle visually but rows never reorder. The browser console shows `[Vue warn]: Duplicate keys found during update`.

### Environment 🌍

- Vue component: [`StorageSection.vue`](https://github.com/ruxailab/RUXAILAB/blob/develop/src/features/navigation/components/navbarSections/StorageSection.vue)
- Git branch: `develop`
- OS: All OS
- Browser: All browsers

### Additional Information ℹ️

The `formatDateLong` function in `src/shared/utils/dateUtils.js` accepts a `locale` parameter and uses `Intl.DateTimeFormat` correctly — the issue here is that `StorageSection.vue` passes a hardcoded `'es'` instead of the user's locale. However, other functions in the same file also have hardcoded locale issues: `formatRelativeTime()` contains hardcoded Spanish strings ("Hoy", "Ayer", "Hace X días" — see #1698), `formatDateTime()` defaults to `'es-ES'`, and `formatDateShort()` defaults to `'en-GB'`. This issue focuses specifically on the `StorageSection.vue` caller.

The duplicate key issue is specific to `StorageSection.vue` because it creates multiple file entries (video, audio, screen, webcam) per task, all sharing the same `task.id`. Other components that use `v-data-table` do not exhibit this issue because they have naturally unique row IDs.

thechillbasu:
## Description
The Storage page (`StorageSection.vue`) had three related bugs: dates were always displayed in Spanish regardless of the user's selected language, clicking table column headers toggled the sort arrow but never reordered the rows, and selecting Portuguese caused all dates to display as hyphens.

**Technical Analysis:**

**Bug 1 — Hardcoded locale:**
The `formatDateLong` call passed a hardcoded `'es'` locale:
```javascript
const date = formatDateLong(answer.date || test.creationDate, 'es')
```
This meant dates always rendered in Spanish (e.g., "18 de febrero de 2026") even when the user had English selected. 

**Bug 2 — Portuguese dates render as hyphens:**
Furthermore, the Portuguese locale code used by the app is `pt_br.json`. Since this has an underscore, `Intl.DateTimeFormat` rejects it as an invalid BCP 47 string, throwing an exception which causes the date formatter to return `-` instead.

**Bug 3 — Sorting not working:**
Two separate issues prevented Vuetify's `v-data-table` from sorting:

1. **Duplicate `id` keys**: Each task can produce multiple file entries (video, audio, screen, webcam), but all shared the same `task.id`. The fallback `self.crypto.randomUUID()` generated a new ID on every computed re-evaluation, causing Vue to emit `Duplicate keys found during update` warnings and preventing Vuetify from tracking and reordering rows.

2. **String-based date column**: The `date` field stored a formatted string like `"18 February 2026"`, which sorted alphabetically (F before J, etc.) rather than chronologically. Vuetify's default comparator cannot sort localized date strings correctly.

## Solution and Reasoning

**Locale fix:**
- Destructured `locale` from `useI18n()` and passed `locale.value` to `formatDateLong()` instead of the hardcoded `'es'`. Dates now automatically follow the user's selected language.
- Added a `replace('_', '-')` step in `dateUtils.js` to normalize the custom `pt_br` locale string into the valid `pt-BR` BCP-47 tag, solving the bug where Portuguese dates rendered as hyphens.

**Sorting fix:**
- **Stable unique IDs**: Removed the `id` from `baseFile` (which previously used `task.id || self.crypto.randomUUID()`) and instead generate a deterministic, unique ID per file entry using the pattern `${testId}_${answerIdx}_${taskIdx}_${fileType}` (e.g., `abc123_0_1_video`). The `baseFile` object is still used to hold shared properties (`studyName`, `date`, `dateFormatted`) and is spread into each file entry via `...baseFile`. Added `item-value="id"` to the `v-data-table` to tell Vuetify to use these IDs for row tracking.

- **Numeric date values**: Split the date into two fields: `date` (raw timestamp via `new Date(rawDate).getTime()` for sorting) and `dateFormatted` (localized display string via `formatDateLong()`). A template slot `#[item.date]` renders `dateFormatted` while Vuetify sorts on the numeric `date`.

- **`custom-key-sort` prop**: Added custom comparison functions for all sortable columns:
  - `studyName`: `localeCompare` with `{ numeric: true }` — so "Test 4" < "Test 10" (natural sort)
  - `date`: numeric subtraction `a - b` (chronological)
  - `size`: numeric subtraction `a - b` (byte-based)

- **`custom-filter` prop**: Since the `date` column now stores a numeric timestamp, the default search filter wouldn't match date text. Added a custom filter that searches against `studyName`, `dateFormatted`, and `type`.

## Changes Made

| File | Change |
|------|--------|
| `dateUtils.js` | Normalize locale by replacing underscores with hyphens (`pt_br` -> `pt-BR`) to prevent `Intl.DateTimeFormat` from throwing |
| `StorageSection.vue` | Use `locale.value` from `useI18n()` instead of hardcoded `'es'` for date formatting |
| `StorageSection.vue` | Move `id` out of `baseFile`, generate stable unique IDs per file using `testId_answerIdx_taskIdx_fileType`; `baseFile` still holds shared props |
| `StorageSection.vue` | Store raw timestamp in `date` field, formatted string in `dateFormatted`, render via template slot |
| `StorageSection.vue` | Add `custom-key-sort` with natural string sort for names and numeric sort for date/size |
| `StorageSection.vue` | Add `custom-filter` to enable search on displayed date strings and study names |
| `StorageSection.vue` | Add `item-value="id"` to `v-data-table` for stable row key tracking |

## Before and After

**Before (Bug)**
Dates show in Spanish ("18 de febrero de 2026") with English selected, and as hyphens (`-`) when Portuguese is selected. Clicking sort arrows does not reorder rows.

https://github.com/user-attachments/assets/1b90e127-84eb-40e6-91d1-a6ae2837cc5d

**After (Fix)**
Dates follow the selected language completely (including Portuguese). Table sorting works correctly for Study Name (natural sort), Date (chronological), and Size (numeric).

https://github.com/user-attachments/assets/1ccd2f9b-338e-4cfa-88dc-f53cf8ba840c

Closes #1712


thechillbasu:
@jvJUCA @KarinePistili Kindly have a look at this!

jvJUCA (approved):
LGTM, Thanks for your contribution

PR #1812, opened 2026-03-06
fix: defer camera initialization in moderated video calls

### LINKED ISSUE #1524: fix(user-test): on moderated test, camera capture starts before entering into a room.
As a moderator, when you enter into a room the camera starts "recording" before the moderator presses on "Start moderator session".

The camera should start once the user presses on "Start moderator session"

rakshityadav1868:
fixes #1524 
## Description
Fix camera initialisation timing in moderated user test sessions. Previously, the camera would start when clicking "Start Test", but it should only start after the moderator clicks "Start Moderated Session".

## AFTER:

https://github.com/user-attachments/assets/05d280a4-1310-4d4c-b403-760c75e700be


mohi-devhub (approved):
@marcgc21 LGTM. Nice focused fix !

PR #1854, opened 2026-03-10
fix(dashboard): calculate usability percentage from real evaluator data

### LINKED ISSUE #1852: Bug: Dashboard usability percentage is hardcoded to 75% instead of real calculation
## Description

The `UsabilityResults.vue` dashboard card always displays **75%** as the general usability metric, regardless of actual evaluation data. This is a data integrity issue — users see fake/placeholder metrics on the test manager dashboard.

## Location

**File:** `src/ux/Heuristic/components/manager/UsabilityResults.vue`  
**Lines:** 82–85

```javascript
const usabilityPercentage = computed(() => {
  // Por ahora devolvemos 75% como solicitado
  // En el futuro esto se calculará basado en las respuestas reales
  return 75
})
```

## Expected Behavior

The usability percentage should be calculated from actual evaluator answers using the same logic as `calcFinalResult()` in `statistics.js` — averaging all evaluators' scores against the maximum possible score (`testOptions`).

## Current Behavior

The value is hardcoded to `75` with a Spanish TODO comment: *"En el futuro esto se calculará basado en las respuestas reales"* (In the future this will be calculated based on real answers).

## Impact

- **Misleading dashboard**: Every test shows "75% usability" with green/warning indicators regardless of actual results
- **Color coding is wrong**: The status text ("Excellent"/"Acceptable"/"Needs improvement") and colors are driven by the hardcoded value, so they also display incorrectly
- **User trust**: Evaluators may make decisions based on a fake metric

## Suggested Fix

Replace the hardcoded value with a real calculation using the Vuex store:

1. Access `store.getters.test` (for `testOptions`) and `store.getters.testAnswerDocument` (for evaluator answers)
2. Use the existing `statistics()` and `calcFinalResult()` functions from `src/ux/Heuristic/utils/statistics.js`
3. Average all evaluators' `result` values, handling the 0-evaluators edge case (return 0, not NaN)

## Screenshots

N/A — the hardcoded `return 75` is visible in the source code.

## Environment

- Vue 3.5 + Vuex 4
- Component: `UsabilityResults.vue`
- Related utility: `src/ux/Heuristic/utils/statistics.js`

KishoreMuruganantham:
## Summary

- Replace hardcoded `return 75` in `UsabilityResults.vue` with a real calculation using the existing `statistics()` utility from `statistics.js`
- Averages all evaluators' `result` values, returning `0` when no evaluators exist (prevents NaN)
- Color coding and status text automatically correct themselves since they already derive from `usabilityPercentage`

Closes #1852

## Before vs After

| Before | After |
|--------|-------|
| Always shows hardcoded 75% regardless of data | Shows real averaged percentage from evaluator scores |

before-after

> The visual component (circular progress, colors, status text) is unchanged — only the underlying data source was fixed from a hardcoded value to a real calculation using `statistics()`

## Test plan

- [ ] Verify dashboard shows 0% usability when no evaluators have submitted answers
- [ ] Verify dashboard shows correct averaged percentage when multiple evaluators have submitted
- [ ] Verify color coding updates correctly: green (≥80%), yellow (≥60%), red (<60%)
- [ ] Verify status text matches the percentage thresholds (Excellent/Acceptable/Needs improvement)


DivyanshuVortex (changes requested):
This PR needs significant refactoring nd so includes several unnecessary changes that appear to have been made and should be removed. Please merge the latest `develop` branch into this PR as well. Thhnx

mohi-devhub (changes requested):
@KishoreMuruganantham 
Hey, thanks for the PR! The core fix replacing the hardcoded 75% is solid. A few things to sort out though:
  - The template access guard in the router got deleted , any logged-in user can now access private templates
  - The emailVerified check in sign-in will lock out all existing unverified accounts with no migration path
  - Verification email failures are silently swallowed (catch {}) , if the email service is down, users get stuck at the verify page forever
  - proceedToNextStep increments globalIndex locally before the Firebase write , roll it back if the write fails
 - qNotApplicable in statistics.js is declared and incremented but never used
 - Two separate imports from the same file in Answer.js , just merge them

This PR touches 70 files for what's a one-line dashboard fix. Would you consider splitting the email verification feature and the router changes into separate PRs? Much easier to review and safer to merge incrementally.

PR #1864, opened 2026-03-11
Fix: preview mode should not save evaluation data (#1856)

### LINKED ISSUE #1856: [🐞 BUG]: preview is NOT a test
### Description 📝

This is useful for ALL evaluation techniques

When we are preparing a test (heuristic, user test, or whichever) we have the 'preview' option, which should allow us to see how what we are preparing will be shown to the evaluators or users.

This seems to work but **counts it as if it were a real test, and that is a very big mistake**

### Link 🔗

https://ruxailab-prod.web.app/heuristic/manager/SYMRMBhLr98Z2AMVNhuX

### Steps to Reproduce 🔄

Go to any technique
On the left menu, click on the option "preview"
Previews, partially or totally, the technique 

### Screenshots 📸

Image

In the image above I show how I have done part of the preview of the heuristic evaluation that I am preparing
It can observed that 4 (from 15) heuristics has been answered, that is 30% (well calculated), as we can observe here:

Image

In answers menu option we have "extrange" information:

at the begining it seems fine: 

Image
nevertheless, the evaluator information is wrong: 

Image
Meanwhile the answers are well saved: 

Image

També apareixo jo com cooperator: 

Image

### Expected Behavior 🤔

As I said before, a **preview is NOT a test**, therefore, a preview should serve so that the person preparing a test can see how it will be displayed but it must NOT save ANY DATA, it cannot appear as if it were a test as it would falsify the data ...

### Actual Behavior 😱

Currently a preview is treated as if it were a real test (at least partially)

### Environment 🌍

-

### Additional Information ℹ️

IMPORTANT: this also happens in user tests.
Next image shows that a user test has been started when it only was a preview !!!

Image

krish01info:
Fixes #1856

## Problem
Preview mode incorrectly saved evaluation answers and counted them as real tests.

## Solution
Added preview-mode guards to prevent Firestore writes when running in preview mode.

## Steps to Verify
1. Create a heuristic evaluation
2. Click Preview
3. Answer questions
4. Return to console

Preview should not create evaluators or store answers.
image
 this is the ss of after  giving the previe it is fixed now


mohi-devhub (changes requested):
@krish01info This conflicts with current `develop` in Firebase config, and the new preview guard treats real `/testview/...` participant sessions as preview, which would skip saving answers. Please rebase and narrow preview detection

PR #1876, opened 2026-03-13
fix(profile): correct deleteAuth payload and remove duplicate deletion logic

### LINKED ISSUE #1872: Bug: Profile account deletion sends wrong deleteAuth payload and double-deletes user
### Issue
The account deletion flow from the profile UI is broken due to two problems:
Incorrect payload sent to deleteAuth
In useDeleteAccount.js, the composable dispatches:
```js
await store.dispatch('deleteAuth', user.uid)
```
However, AuthController.deleteAuth() expects a payload like:
`{ user, password }`
Since only a UID string is passed, the controller destructures the payload and throws:
```js
const { user, password } = payload
if (!user) throw new Error('No user provided')
```
This causes the deletion flow to fail.
### Duplicate Deletion Logic
User deletion is executed in multiple layers.
In `useDeleteAccount.js`:
- `deleteDoc(userDocRef)` deletes the Firestore document
- `store.dispatch('deleteAuth', ...)` is called
- `deleteUser(user)` is called again

But `AuthController.deleteAuth()` already performs:
- `user.delete()` for Firebase Auth removal
- `deleteUserData(user.uid)` for backend cleanup
This duplicates deletion logic across layers.
### Steps to Reproduce
1. Sign in.
2. Navigate to profile settings.
3. Trigger Delete Account.
4. Observe deleteAuth is dispatched with user.uid instead of { user, password }.
### Expected Behavior
Account deletion should:
- reauthenticate the user
- pass the correct payload to the controller
- delete user data once
- sign the user out after successful deletion

### Screenshot
Image

prakriti-shrestha:
## What does this PR do?
Fixes two bugs in the account deletion flow from the profile UI that caused deletion to fail silently and attempt multiple redundant operations.

## Related Issue
Closes #1872

## Root Cause Analysis
**Bug 1 - Wrong payload to 'deleteAuth'**
'useDeleteAccount.js' was dispatching:
```js
await store.dispatch('deleteAuth', user.uid) // raw UID string
```
But 'AuthController.deleteAuth()' destructures its payload as an object:
```js
const { user, password } = payload
if (!user) throw new Error('No user provided') // always threw
```
This caused every deletion attempt to fail immediately.

**Bug 2 - Triple deletion**
The composable was executing deletion in three separate place:
- 'deleteDoc(userDocRef)' - manual Firestore deletion
- 'store.dispatch('deleteAuth', ...)' - controller calls `user.delete()` + `deleteUserData()`
- `deleteUser(user)` — redundant Firebase Auth deletion

**Bug 3 - Double reauthentication**
The composable reauthenticates the user before calling `deleteAccount()`, but `AuthController.deleteAuth()` was also reauthenticating internally, causing redundant popup/credential prompts.

## Changes Made
**`useDeleteAccount.js`**
- Fixed dispatch payload from `user.uid` to `{ user, password }`
- Added `password` parameter to `deleteAccount()` and passed it through from both `handleDeleteAccount` and `handleDeleteConfirmText`
- Removed `deleteDoc(userDocRef)` - backend cleanup is handled by controller
- Removed redundant `deleteUser(user)` call
- Removed unused imports: `deleteUser`, `doc`, `deleteDoc`, `db`

**`AuthController.js`**
- Removed internal reauthentication block - composable handles this before dispatching, so controller only needs to execute deletion
- Removed unused imports: `reauthenticateWithCredential`, `reauthenticateWithPopup`, `EmailAuthProvider`

## Deletion Flow After Fix
```
handleDeleteAccount / handleDeleteConfirmText
  → reauthenticate (once, in composable)
  → store.dispatch('deleteAuth', { user, password })
      → user.delete()           (Firebase Auth removal)
      → deleteUserData(user.uid) (backend cleanup, best-effort)
  → showSuccess()
  → signOut() → redirect to /signin
```

## Testing
- [x] Email account deletion: correct password → account deleted, redirected to /signin
- [x] Email account deletion: wrong password → shows `errors.incorrectPassword`
- [x] Google account deletion: popup accepted → account deleted, redirected to /signin  
- [x] Google account deletion: popup closed → shows `errors.authenticationCancelled`
- [x] Firebase Auth console: user removed after deletion
- [x] Firestore console: user document removed after deletion

## Demo
**Before:** Deletion fails silently, user remains in Firebase Auth

**After:**
Email signup deletion flow
https://github.com/user-attachments/assets/ed8e3304-1a51-43dc-85e8-32c1467d687b

Google account signup deletion flow

https://github.com/user-attachments/assets/4c174364-3be2-4276-87ff-4938fb4b25d2

### DIFF
--- a/src/features/auth/composables/useDeleteAccount.js
+++ b/src/features/auth/composables/useDeleteAccount.js
@@ -6,10 +6,7 @@ import {
   EmailAuthProvider,
   reauthenticateWithPopup,
   GoogleAuthProvider,
-  deleteUser,
 } from 'firebase/auth'
-import { doc, deleteDoc } from 'firebase/firestore'
-import { db } from '@/app/plugins/firebase'
 import { showError, showSuccess } from '../../../shared/utils/toast'
 
 const SHARED_ERRORS = {
@@ -44,18 +41,9 @@ export function useDeleteAccount() {
     showError(messageKey)
   }
 
-  const deleteAccount = async (user) => {
+  const deleteAccount = async (user, password) => {
     try {
-      const userDocRef = doc(db, 'users', user.uid)
-      await deleteDoc(userDocRef)
-
-      try {
-        await store.dispatch('deleteAuth', user.uid)
-      } catch (storeError) {
-        return storeError
-      }
-
-      await deleteUser(user)
+      await store.dispatch('deleteAuth', { user, password })
 
       showSuccess('profile.accountDeletedSuccess')
 
@@ -63,15 +51,6 @@ export function useDeleteAccount() {
     } catch (error) {
       if (error.code === 'permission-denied') {
         showError('profile.permissionDenied')
-      } else if (error.code === 'not-found') {
-        try {
-          await deleteUser(user)
-          showSuccess('profile.accountDeletedSuccess')
-          await signOut()
-        } catch (authError) {
-          showError('profile.accountDeletionFailed')
-          throw authError
-        }
       } else {
         showError('profile.accountDeletionFailed')
         throw error
@@ -112,7 +91,7 @@ export function useDeleteAccount() {
     try {
       isDeleting.value = true
       await reauthenticateWithPopup(user, new GoogleAuthProvider())
-      await deleteAccount(user)
+      await deleteAccount(user, null)
     } catch (error) {
       handleAuthError(error, GOOGLE_ERRORS)
     } finally {
@@ -140,7 +119,7 @@ export function useDeleteAccount() {
       const cred = EmailAuthProvider.credential(user.email, userPassword.value)
       await reauthenticateWithCredential(user, cred)
 
-      await deleteAccount(user)
+      await deleteAccount(user, userPassword.value)
     } catch (error) {
       handleAuthError(error, EMAIL_ERRORS)
     } finally {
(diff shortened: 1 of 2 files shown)


prakriti-shrestha:
Hi @marcgc21 I have opened this fix for #1872 . Happy to make any changes based on your feedback!

Karrrthik7:
Since I originally opened this issue and had already drafted a fix, I want to raise one point regarding this change.
The PR removes the reauthentication logic from `AuthController.deleteAuth()` and shifts it entirely to the composable. This means the controller now assumes the caller has already reauthenticated the user. Since deleteAuth is dispatched through the store, could this introduce a dependency on caller behavior if it is used elsewhere without prior reauthentication?
Just wanted to clarify this before the change gets merged.

prakriti-shrestha:
Hi @Karrrthik7 that is a fair and valid point. You are right that removing reauthentication from the controller creates an implicit precondition on the caller. 

Right now 'deleteAuth' is only dispatched from 'useDeleteAccount.js', which always reauthenticates before calling it. So the behaviour is correct for the existing use case. However, I agree that if 'deleteAuth' were called from another context without prior reauthentication, it would fail unexpectedly.

To make this contract explicit and safe, I can add a JSDoc comment to 'AuthController.deleteAuth()' clearly documenting the precondition:
 * Deletes the authenticated user account and cleans up backend data.
 * @precondition Caller must reauthenticate the user before calling this method.
 * @param {Object} payload
 * @param {Object} payload.user - Firebase auth user (already reauthenticated)

This way any future caller is explicitly warned. Would that address your concern, or would you prefer the reauthentication logic stays in the controller as a safety net? Happy to revert that part if Marc prefers the controller to be self-contained.

KarinePistili (approved):

PR #1879, opened 2026-03-13
fix(heuristic): disable submit button until evaluation progress reaches 100%

### LINKED ISSUE #1873: [🐞 BUG]: Heuristic evaluation submit button is never disabled due to incorrect progress reference
## Description

In the heuristic evaluation view, the submit Floating Action Button (FAB) is intended to be disabled until the evaluator has answered all questions.

However, the button is always enabled due to an incorrect variable reference in the template.

The `:disabled` attribute currently compares a function reference instead of the reactive progress value.

### Affected file

src/views/HeuristicTestView.vue

Approximate line: 543

### Current implementation

```vue
:disabled="calculateProgress < 100
```
Here calculateProgress is a function (defined later in the file), not the reactive ref that stores the computed progress value.

In Vue templates, a function reference is always truthy, therefore:
```vue   
calculateProgress < 100   
```

always evaluates to false, which means the submit button is never disabled.

### Correct behavior

The template should reference the reactive progress variable:

```vue 
 :disabled="calculatedProgress < 100" 
 ```

which is already used correctly in other progress checks within the same file.

### Link 🔗

https://github.com/ruxailab/RUXAILAB/blob/develop/src/ux/Heuristic/views/HeuristicTestView.vue

### Steps to Reproduce 🔄

1. Start a heuristic evaluation study.
2. Navigate to the evaluation interface.
3. Do not answer all evaluation questions.
4. Observe the floating submit button.

The submit button remains enabled even when the evaluation progress is incomplete.

### Screenshots 

Image

Image

### Expected Behavior 

The submit button should remain disabled until the evaluator has completed all required evaluation questions.

### Actual Behavior 

The submit button is always enabled regardless of the evaluation progress because the template compares a function reference instead of the computed progress value.

### Environment 

Vue component: src/ux/Heuristic/views/HeuristicTestView.vue
Branch: main
Browser: Chrome
OS: macOS

### Additional Information 

The correct reactive variable `calculatedProgress` already exists in the same file and is used in other progress checks. Replacing the incorrect reference should resolve the issue.

aaradhychinche-alt:
## Related Issue
Fixes #1873

## Problem

During heuristic evaluations, the **Submit** button is never properly disabled based on the evaluation progress.

The template currently checks:

:disabled="calculateProgress < 100"

However, `calculateProgress` is a function reference instead of the actual computed progress value. Because of this, the condition does not behave as expected and the submit button can remain enabled even when the evaluation is incomplete.

This can allow users to submit evaluations before all questions are answered.

## Root Cause

The template references the progress calculation function instead of the computed progress value.

## Solution

Updated the submit button condition to use the correct reactive progress value instead of the function reference.

This ensures:

- Submit button remains **disabled until progress reaches 100%**
- Users cannot submit incomplete heuristic evaluations
- Progress logic works as intended.

## UI Screenshots
Screenshot 2026-03-14 at 4 19 50 AM

### Before Fix
Submit button could remain enabled even when evaluation progress was incomplete.

### After Fix
Submit button stays disabled until all heuristic questions are answered and progress reaches 100%.

## Testing

Tested locally by:

1. Creating a heuristic study
2. Adding multiple heuristic questions
3. Starting the evaluation
4. Verifying that:
   - Submit button stays disabled when progress < 100%
   - Submit button becomes enabled only after all questions are answered

## Notes

This change only updates the progress reference used in the template and does not modify existing evaluation logic.


aaradhychinche-alt:
@marcgc21  @KarinePistili  can you please look at this and other issues i raised !!

PR #1880, opened 2026-03-14
fix(profile): correct deleteAuth payload and remove duplicate deletion

### LINKED ISSUE #1872: Bug: Profile account deletion sends wrong deleteAuth payload and double-deletes user

Karrrthik7:
Fixes #1872
### Summary
Fixes the profile account deletion flow by sending the correct payload to the `deleteAuth` store action and removing redundant deletion logic from the composable.

### Problem
The profile deletion flow dispatched `deleteAuth` with a UID string:
```js
await store.dispatch('deleteAuth', user.uid)
```
However, the controller expects the payload in the following format:
`{ user, password }`
Because of this mismatch, the controller failed with: `Error: No user provided.`
Additionally, the composable also deleted the Firestore document and Firebase user locally before dispatching deleteAuth, which could lead to duplicate deletion attempts and inconsistent cleanup.
### Changes
- Pass `{user, password}` payload to `deleteAuth`
- Remove local Firestore document deletion from the composable
- Remove redundant `deleteUser()` call
- Keep account deletion centralized in the auth controller
### Result
The deletion flow now follows a single responsibility path. The profile UI calls `deleteAuth` through the store, which triggers `AuthController.deleteAuth`. The controller then handles the Firebase Auth deletion and backend user data cleanup, avoiding duplicate deletion logic.
### Before Changes
https://github.com/user-attachments/assets/921ff145-4192-4845-bcda-8c73d6797e26
### After Changes
https://github.com/user-attachments/assets/caf72657-b948-4985-83ef-64b6e20620a7

### DIFF
--- a/src/features/auth/composables/useDeleteAccount.js
+++ b/src/features/auth/composables/useDeleteAccount.js
@@ -44,51 +44,22 @@ export function useDeleteAccount() {
     showError(messageKey)
   }
 
-  const deleteAccount = async (user) => {
+  const deleteAccount = async (user, password) => {
     try {
-      const userDocRef = doc(db, 'users', user.uid)
-      await deleteDoc(userDocRef)
-
-      try {
-        await store.dispatch('deleteAuth', user.uid)
-      } catch (storeError) {
-        return storeError
-      }
-
-      await deleteUser(user)
+      await store.dispatch('deleteAuth', { user, password })
 
       showSuccess('profile.accountDeletedSuccess')
-
       await signOut()
     } catch (error) {
-      if (error.code === 'permission-denied') {
-        showError('profile.permissionDenied')
-      } else if (error.code === 'not-found') {
-        try {
-          await deleteUser(user)
-          showSuccess('profile.accountDeletedSuccess')
-          await signOut()
-        } catch (authError) {
-          showError('profile.accountDeletionFailed')
-          throw authError
-        }
-      } else {
-        showError('profile.accountDeletionFailed')
-        throw error
-      }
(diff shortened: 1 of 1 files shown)


Karrrthik7:
cc @marcgc21

Karrrthik7:
cc @marcgc21

PR #1923, opened 2026-03-18
Feat(superadmin): Unified User & Study Management Dashboard (#1215)

### LINKED ISSUE #1215: feat(sa): as a super-admin I want a table to see all users
As a super-admin I would like to see information related to the users logged into the system. The columns might be:

- Name
- Email
- Number of studies created
- Total media size

Cizelle:
This PR consolidates the fragmented SuperAdmin experience into a single, high-efficiency tabbed view.

**Key Enhancements:**
- Unified UI: Replaced separate routes with a v-tabs interface in SuperAdminView.vue for seamless switching between Users and Studies.
- State Reactivity: Synchronized User.js and Study.js store modules to ensure real-time table updates upon deletion or role changes.
- UX Polish: Implemented formatMediaSize for storage data and automated search-clearing on tab navigation to prevent filtered "ghost" results.
- Routing Fix: Resolved the "blank page" bug by simplifying the superAdmin.js route structure and updating the global redirect() logic for Level 0/4 access.

Fixes #1215.
Screenshot 2026-03-16 224804
Screenshot 2026-03-17 230615


Cizelle:
Hi @marcgc21 , I've noticed the CI is failing on EmailController.spec.js. It looks like an Axios ESM syntax error (import statement outside a module) which seems unrelated to the SuperAdmin UI and routing changes in this PR.

Since this is a separate module, I'll leave this for now and focus on the project proposals, but let me know if you'd like me to look into the Jest config!

DivyanshuVortex (changes requested):
Hii @Cizelle , thanks for the pr , but the issue is in the router: the `beforeEach` guard was simplified to just return `next()`. This completely removes all authentication, authorization, and email verification checks, making route public /superadmin.

Also, the changes made to the screen recording stop logic in ScreeenRecorder.vue and UserTestView.vue . Can you please explain why these recording changes were implemented this way?

Cizelle:
Hi @DivyanshuVortex , thanks for the PR review
About the router - thanks for pointing that out. It was a temporary bypass for local testing that I forgot to revert. I'll get the secure version back in place, and as for the ScreenRecorder changes - I'm actually glad you brought that up. I’m working on a commit to ensure the task-save doesn't trigger until the recording upload is 100% verified (to avoid the race condition). I'll push an update covering both the security fix and the recording stability soon, do check that out too!

mohi-devhub (changes requested):
@Cizelle `ScreenRecorder.vue` uses `watch()` but does not import it inside `<script setup>`, so `stopRecording()` will fail at runtime. Also, `/superadmin` allows level 4, but the UI only labels/edits levels 0 and 1

KarinePistili:
I’m rejecting this PR for now because it mixes unrelated changes (the PR is about SuperAdmin but also modifies recording/media flows), the UI changes feel inconsistent, and it introduces risky auth/navigation changes while the recording flow still appears to save before uploads finish, which may cause missing media data.

PR #1924, opened 2026-03-19
Accessibility: Add aria-labels to icon-only buttons

Ujjwal-Gowda:
This PR improves the accessibility of the platform by adding descriptive aria-label attributes to interactive elements that previously only contained icons it helps users relying on screen readers or other assistive technologies to understand the purpose of these buttons. And it also complies with WCAG(web content accessibility guidelines ) which just mean its helpfull for people with disabilitys to use and multi language support
Screenshot from 2026-03-19 20-09-11

### DIFF
--- a/src/features/language/components/LocaleChanger.vue
+++ b/src/features/language/components/LocaleChanger.vue
@@ -21,7 +21,13 @@
     >
       <v-menu>
         <template #activator="{ props }">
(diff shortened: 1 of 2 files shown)


DivyanshuVortex (approved):
LGTM .. Thanks for the contribution

PR #1929, opened 2026-03-20
chore: add accessible names to buttons & progress indicators

### LINKED ISSUE #1928: Accessibility: Missing Accessible Names
Some buttons and progress indicators do not have accessible names.

**Impact:**
Screen readers announce them as just “button” or “progressbar”, making them unusable.

**Affected Elements:**

* `v-btn` icon buttons (no `aria-label` or text)
* `v-progress-circular` with `role="progressbar"` (no label)

**Fix:**

* Add `aria-label` or `aria-labelledby` to buttons
* Add accessible name to progressbar (e.g., `aria-label="Loading progress"`)

**WCAG:** 4.1.2 (Name, Role, Value)

antaripdebgupta:
This PR improves accessibility by adding missing accessible names to icon-only buttons and progress indicators.

**Changes:**

* Added `aria-label` to all `v-btn` icon buttons
* Added accessible name (`aria-label` / `aria-labelledby`) to `v-progress-circular`
* Ensures screen readers announce meaningful labels instead of generic terms

**Related Issue**
Fixes: #1928 

**Impact:**
Improves usability for screen reader users and aligns with WCAG 4.1.2 (Name, Role, Value)

**Image**
accessbility


Patrick-Augusto:
@marcgc21
The CI is failing because EmailController.js imports axios, and Jest is not handling the current ESM version of axios in the test environment.

Failing error:
SyntaxError: Cannot use import statement outside a module

How to fix:

Mock axios in EmailController.spec.js if the test does not need the real HTTP call.
Or update Jest config to transform axios.
Or use an axios version compatible with the current Jest setup.
Recommended fix: mock axios in the unit test, since it is the simplest and safest option for this PR.

Issue #1990, opened 2026-03-29
Feature: Multimodal Input Modules for Text Sentiment and Audio Emotion Recognition - GSOC-2026


HITESH-S-P:
**Labels:** `enhancement` `gsoc-2026` `multimodal` 
**Depends on:** #1988 

---

### What it should do

Take a string of text - a participant's transcribed comment, their answer to a task question, and return a `SentimentOutput` with `modality: "text"`.

### Intent detection (optional but valuable)

If time allows, it would be useful to also detect *intent* from the text and not just what emotion the participant expressed, but whether they're making a complaint, a suggestion, asking a question, or describing confusion. This doesn't need a separate model, a simple keyword/pattern approach is fine for now. The intent label can feed into the usability mapping the same way an emotion label does.

---

### What it should do

Take a path to an audio file (WAV or MP3, short clip from a usability session recording) and return a `SentimentOutput` with `modality: "audio"`.

---

Issue #1988, opened 2026-03-29
Feature: Unified Sentiment Processing Pipeline with Confidence Estimation and Usability-Oriented Mapping -GSOC-2026


HITESH-S-P:
<html>
<body>
<html><head></head><body><h1>Unified Sentiment Processing Pipeline with Confidence Estimation and Usability-Oriented Mapping - (Phase-1)</h1>
<p><strong>Labels:</strong> <code>enhancement</code> <code>gsoc-2026</code> <code>sentiment analysis</code>
<hr>
<h2>What needs to be built</h2>

<h3>1. Standard output schema</h3>
<p>
All modules (text, audio, facial) should return results in a common format. Each output must include:
</p>
<ul>
  <li><code>emotion</code> - predicted label</li>
  <li><code>confidence</code> - score between 0 and 1</li>
  <li><code>usability_label</code> - mapped UX-focused category (e.g. frustration, satisfaction)</li>
  <li><code>modality</code> - source (<code>text</code>, <code>audio</code>, <code>face</code>, <code>fused</code>)</li>
  <li><code>raw_scores</code> - optional probabilities for all classes</li>
</ul>

<h3>2. Confidence extraction</h3>
<p>
Model outputs (logits) should be converted into probabilities using <strong>softmax</strong>.
The confidence score must always be the probability of the top prediction and not raw values.
</p>
<p>
This will be important later when combining multiple modalities.
</p>

<h3>3. Emotion to usability mapping</h3>
<p>
We need a simple mapping from generic emotion labels to usability-focused categories.
</p>
<ul>
  <li>Clear and readable (dictionary-based)</li>
  <li>Easy to extend</li>
  <li>Separate from core logic</li>
</ul>

<table>
  <thead>
    <tr>
      <th>Emotion</th>
      <th>Usability</th>
    </tr>
  </thead>
  <tbody>
    <tr><td>anger, disgust, sadness</td><td>frustration</td></tr>
    <tr><td>fear</td><td>cognitive_load</td></tr>
    <tr><td>joy</td><td>satisfaction</td></tr>
    <tr><td>surprise</td><td>engagement</td></tr>
    <tr><td>neutral</td><td>neutral</td></tr>
    <tr><td>confusion</td><td>confusion</td></tr>
  </tbody>
</table>

<hr>
<h2>Expected output format</h2>
<pre><code class="language-json">{
  "emotion": "confusion",
  "confidence": 0.81,
  "usability_label": "cognitive_load",
[...]

PR #1993, opened 2026-03-29
feat: weighted late fusion with modality contribution tracking and explainability output

### LINKED ISSUE #1992: Feature: Multimodal Sentiment Fusion with Explainable Modality Contributions -GSOC-2026
# Feature: Multimodal Sentiment Fusion with Explainable Modality Contributions -GSOC-2026 

**Labels:** `enhancement` `gsoc-2026` `explainability` `core`
**Depends on:** #1988 , #1990 

---

### 1. Fusion module

The approach is **weighted late fusion**. "Late" means each modality runs its full pipeline and produces a complete prediction before fusion happens — we're combining at the prediction level, not at the feature level. This is the right choice for the current architecture because the modalities are processed independently and may not all be available in every session (a session might have text and audio but no facial data, for example).

The fusion works like this:

1. Collect predictions from all available modalities
2. For each modality, weight its raw probability scores by the modality's assigned weight
3. Sum the weighted scores across all modalities for each emotion class
4. Normalize so the final scores form a valid probability distribution
5. Pick the top emotion as the final prediction

The default weights should reflect the general reliability ordering for usability contexts:

```
face:  0.40
audio: 0.35
text:  0.25
```

These weights are a starting point, not a fixed constraint. The module should accept custom weights at instantiation so they can be tuned based on the modalities available in a given study or on empirical results.

The fusion module must handle partial modality sets gracefully - if only text and audio are available (no facial module), it should renormalize the remaining weights to sum to 1.0 and proceed.
---

HITESH-S-P:
**Closes** #1992
**Branch:** `feature/multimodal-fusion-explainability` 
**Depends on:** #1988 , #1990 
**Labels:** `enhancement` `gsoc-2026` `core`

---

## Outcomes of this PR

This is the core of the multimodal sentiment engine - the piece that takes predictions from text, audio, and (eventually) facial modules and combines them into a single explainable output and Fixes #1992 .

Two files:

- `fusion/multimodal_fusion.py` — weighted late fusion over per-modality `SentimentOutput` objects
- `explainability/contributions.py` — computes normalized modality contribution scores

---

## How contributions are calculated

After fusion, the contribution of each modality is:

```
contrib(m) = weight(m) × confidence(m)
             ─────────────────────────────
             Σ [ weight(i) × confidence(i) ]
```

This weights each modality by both its assigned importance and how confident it actually was on this particular input. A modality that was uncertain contributes less to the explanation even if it had a high assigned weight — which I think is the right behavior for a research setting where you want to know when a signal was strong.

The contributions sum to 1.0 and live in the `contributions` field of the returned `SentimentOutput`.

---
## Testing

```bash
clone repo
cd sentiment-engine
pip install -r requirements.txt
pytest tests/test_fusion.py -v
```

Test coverage includes:
- Three modalities agree → high confidence, contributions roughly proportional to weights
- Three modalities disagree → moderate confidence, contributions reflect per-modality confidence × weight
- Two modalities only → renormalization works, contributions still sum to 1.0
- One modality → fusion degrades gracefully (returns that modality's output with `modality: "fused"`)
- Contributions sum-to-one invariant across all cases
- Custom weights are respected in both fusion and contribution output'

Output:
Screenshot 2026-03-29 215759

---

```json
{
  "emotion": "anger",
  "confidence": 0.6892,
  "usability_label": "frustration",
  "modality": "fused",
  "raw_scores": {
    "anger": 0.6892,
    "neutral": 0.2184,
    "joy": 0.0924
  },
  "contributions": {
    "face": 0.4331,
    "audio": 0.3419,
    "text": 0.2250
  }
}
```

The text module disagreed (predicted neutral) but had lower confidence, so its contribution is smallest and anger still wins. That's the behavior we want.

---

## Files changed

```
fusion/multimodal_fusion.py           
explainability/contributions.py       
tests/test_fusion.py        
```

---

This PR is part of my GSoC 2026 application for the Multimodal Sentiment Analysis Engine project.

### DIFF
--- a/sentiment-engine/.gitignore
+++ b/sentiment-engine/.gitignore
@@ -1,0 +1,3 @@
+venv
+.pytest_cache
+__pycache__/
--- a/sentiment-engine/README.md
+++ b/sentiment-engine/README.md
@@ -1,0 +1,186 @@
+# Unified Sentiment Processing Pipeline
+
+Branch: `feature/unified-sentiment-pipeline`
+
+This branch introduces the core processing layer for the RUXAILAB multimodal sentiment analysis engine. It establishes a shared output schema, a confidence extraction mechanism, and a usability-oriented emotion mapping layer that all downstream modules (text, audio, face) will depend on.
+
+---
+
+## Why this exists
+
+Sentiment models whether they analyze text, audio, or facial expressions, tend to return predictions in whatever format the underlying library spits out. When you have one modality that's fine. When you're trying to combine three, it becomes a mess fast.
+
+This layer exists to enforce a single output contract across the entire pipeline. It also handles two things the raw model outputs don't give you: a normalized confidence score (softmax over raw logits) and a mapping from general emotion labels to the usability-specific categories that RUXAILAB's research taxonomy uses.
+
+---
+
+## Structure
+
+```
+schemas/
+  output_schema.py       — Pydantic model defining the standard output structure
+
+utils/
+  emotion_mapper.py      — Maps emotion labels to usability-oriented categories
+
+pipeline/
+  sentiment_processor.py — Core processor: takes raw scores, returns SentimentOutput
+
+tests/
+  test_sentiment_processor.py
+  test_emotion_mapper.py
+```
+
+---
+
+## Quick start
+
+```bash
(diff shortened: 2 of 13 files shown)


HITESH-S-P:
Hi, just following up on this PR. Would appreciate a review when you have time

PR #2012, opened 2026-03-31
feat(transcriptions): add utility functions for transcript processing

Cloxy7:
Fixes #2011 

## Summary
This PR introduces utility functions to simplify extraction and preparation of transcription data for downstream processing.

## Changes
- Added function to extract full transcript text (moderator + evaluator)
- Added function to retrieve all segments
- Added function to prepare segments for NLP processing

These utilities provide a clean and reusable interface for working with transcription data and enable future NLP-based analysis (intent detection, sentiment analysis, etc.).

Screenshot 2026-03-31 093647

### DIFF
--- a/src/ai/transcriptions/transcription.utils.js
+++ b/src/ai/transcriptions/transcription.utils.js
@@ -1,0 +1,37 @@
+/**
+ * Extract full transcript text (moderator + evaluator)
+ */
+export function getAllTranscriptText(transcription) {
+  if (!transcription) return ''
+
+  const moderatorText = transcription.moderator?.transcript || ''
+  const evaluatorText = transcription.evaluator?.transcript || ''
+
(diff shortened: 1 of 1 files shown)


Patrick-Augusto:
@marcgc21 
The CI is failing because EmailController.js imports axios, and Jest is not handling the current ESM version of axios in the test environment.

Failing error:
SyntaxError: Cannot use import statement outside a module

How to fix:

Mock axios in EmailController.spec.js if the test does not need the real HTTP call.
Or update Jest config to transform axios.
Or use an axios version compatible with the current Jest setup.
Recommended fix: mock axios in the unit test, since it is the simplest and safest option for this PR.

marcgc21:
We are going to close this PR because no issue was attached.

PR #2027, opened 2026-04-08
feat: text sentiment and audio emotion modules with MFCC-based feature extraction

### LINKED ISSUE #1990: Feature: Multimodal Input Modules for Text Sentiment and Audio Emotion Recognition - GSOC-2026

HITESH-S-P:
# feat: text sentiment and audio emotion modules with MFCC-based feature extraction

**Closes** #1990 
**Branch:** `feature/multimodal-input-modules`
**Depends on:** #1988 
**Labels:** `enhancement` `gsoc-2026`

---
Pipeline of this project
Screenshot 2026-03-30 223716

---
## Outcomes of this PR

Builds the two input modules described and hence Fixes #1990  - text sentiment analysis and audio emotion recognition, along with the audio preprocessing pipeline that feeds the audio classifier.

Both modules return `SentimentOutput` objects via `SentimentProcessor`, so they plug directly into the schema established in Issue #1988 without any glue code.

---

## Text module (`modules/text_sentiment.py`)

Uses `j-hartmann/emotion-english-distilroberta-base` from Hugging Face. The model is loaded once at class init using the `transformers` pipeline API with `top_k=None`, which returns probabilities for all 7 emotion classes. Those scores are passed straight to `SentimentProcessor.process()`.
The model is loaded lazily — only when `TextSentimentAnalyzer` is first instantiated, not at import time. This matters because importing the module in tests shouldn't trigger a 300MB model download.
Empty string input returns a neutral output with 0.0 confidence rather than raising. Usability sessions produce noisy transcripts; the module shouldn't crash on an empty turn.

---

## Audio feature extractor (`preprocessing/audio_features.py`)

Pitch extraction uses `librosa.piptrack()` with a voiced-frames filter (`magnitude > 0`). If no voiced frames are detected (e.g. silence), pitch defaults to 0.0 rather than NaN.
The feature vector is 16 values total (13 MFCCs + 3 scalar features). This is what the audio classifier trains and predicts on.

---

## Audio emotion classifier (`modules/audio_emotion.py`)

SVM with RBF kernel and `probability=True`. The scaler and classifier are fit together in `.train(X, y)`. Predictions go through `SentimentProcessor` as with text.

When the classifier hasn't been trained, `.analyze()` falls back to a rule-based heuristic:
- High energy + high pitch → lean toward `anger`
- Very low energy → lean toward `sadness`
- Moderate energy + mid pitch → lean toward `neutral`

This fallback exists so the module is functional out of the box for testing, before we have the annotated usability study dataset to actually train on. It's not meant to be a real classifier — just something that returns the right shape of output.

The `.train()` method accepts numpy arrays directly, which means it can be called with features extracted from a labeled audio dataset in a notebook or training script without needing to go through the file-based `analyze()` path.

---

The audio tests use synthetic signals (pure tones and white noise generated with numpy) - no real audio files are committed to the repo.

The text tests use hardcoded strings that should have predictable outputs:

```
"That worked perfectly, exactly what I expected"
→ emotion: joy, usability_label: satisfaction

"This button doesn't work and I can't find the submit option"
→ intent override → usability_label: frustration
```

---

## Files changed

```
modules/text_sentiment.py       
modules/audio_emotion.py          
preprocessing/audio_features.py   
requirements.txt             
```

---

## Known limitations

The text model (`distilroberta-base`) is English-only. Multilingual support will need a separate model - not in scope for this issue.

The audio classifier is untrained until you call `.train()` with real data. The heuristic fallback is coarse. Once we have annotated recordings from usability sessions, the classifier will be much more useful.

Audio files need to be loadable by `librosa` - WAV and MP3 work, but some FLAC encodings have caused issues in my testing. I've added a try/except around the load call that logs a warning and returns a neutral output rather than crashing.

---

This PR is part of my GSoC 2026 application for the Multimodal Sentiment Analysis Engine project.

### DIFF
--- a/sentiment-engine/.gitignore
+++ b/sentiment-engine/.gitignore
@@ -1,0 +1,3 @@
+venv
+.pytest_cache
+__pycache__/
--- a/sentiment-engine/README.md
+++ b/sentiment-engine/README.md
@@ -6,7 +6,7 @@ This branch adds the text sentiment and audio emotion modules to the RUXAILAB se
 
 ---
 
-## What's in here
+## What is there in this update
 
 ```
 modules/
--- a/sentiment-engine/README.md
+++ b/sentiment-engine/README.md
@@ -1,0 +1,186 @@
+# Unified Sentiment Processing Pipeline
+
+Branch: `feature/unified-sentiment-pipeline`
+
+This branch introduces the core processing layer for the RUXAILAB multimodal sentiment analysis engine. It establishes a shared output schema, a confidence extraction mechanism, and a usability-oriented emotion mapping layer that all downstream modules (text, audio, face) will depend on.
+
+---
+
+## Why this exists
+
+Sentiment models whether they analyze text, audio, or facial expressions, tend to return predictions in whatever format the underlying library spits out. When you have one modality that's fine. When you're trying to combine three, it becomes a mess fast.
+
+This layer exists to enforce a single output contract across the entire pipeline. It also handles two things the raw model outputs don't give you: a normalized confidence score (softmax over raw logits) and a mapping from general emotion labels to the usability-specific categories that RUXAILAB's research taxonomy uses.
+
+---
+
+## Structure
+
+```
+schemas/
+  output_schema.py       — Pydantic model defining the standard output structure
+
+utils/
+  emotion_mapper.py      — Maps emotion labels to usability-oriented categories
+
+pipeline/
+  sentiment_processor.py — Core processor: takes raw scores, returns SentimentOutput
+
+tests/
+  test_sentiment_processor.py
+  test_emotion_mapper.py
+```
+
+---
+
+## Quick start
+
+```bash
+git clone https://github.com/HITESH-S-P/multimodal-sentiment-engine.git
+cd sentiment-engine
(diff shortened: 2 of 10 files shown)


HITESH-S-P:
> 🚫 **Too many open Pull Requests**
> 
> You already have **3 open PRs** in this repository.
> 
> Please finish or close existing PRs before opening new ones.

Currently there is only 2 open PRs from my side not 3, why is this blocking the 3rd PR.

PR #2127, opened 2026-06-19
fix: prevent user studies from disappearing

### LINKED ISSUE #2126: [🐞 BUG]: User studies disappear — myTests overwritten by stale user doc
## Description 📝

A user's studies can suddenly disappear from their account. The studies still exist in Firestore (`tests` collection) — what gets wiped is the user doc's `myTests` field, which links a user to their studies.

Cause: the `user` object in the store is loaded once at login and never refreshed, but new studies are added to `myTests` **server-side** by the `onTestCreate` Cloud Function. Several actions then rewrite the **whole** user doc from the stale browser copy via `User.toFirestore()` (which always includes `myTests`), overwriting the server's correct value. The main trigger is marking a notification as read.

## Link 🔗

Notifications panel / bell (e.g. `/admin?section=notifications`)

## Steps to Reproduce 🔄

1. Log in with existing studies.
2. Create a new study (added to `myTests` server-side; browser copy not updated).
3. Click a notification, or "Mark all as read".
4. Refresh — studies are gone.

## Screenshots 📸

Reported on Discord: study docs still in Firestore, but user doc `myTests` is empty.

## Expected Behavior 🤔

Marking a notification read should only update notifications, never touch `myTests`.

## Actual Behavior 😱

The whole user doc is overwritten from a stale copy, unlinking the user's studies.

## Environment 🌍

- `src/features/auth/controllers/UserController.js` — `markNotificationAsRead` (`new User(payload.user)` → `toFirestore()`)
- `src/features/auth/models/UserModel.js` — `toFirestore()` always writes `myTests`
- `functions/src/triggers/onTestCreate.js` — adds studies to `myTests` server-side
- Branch: `develop`

## Additional Information ℹ️

Same overwrite pattern affects accepting a collaboration invite and deleting a study. Fix: use field-scoped `updateDoc` so each action only updates its own field (don't round-trip `myTests`). Affected users are recoverable since study docs still exist.

mohi-devhub:
## Description

Users reported that studies they created would suddenly disappear from their account, even though the study documents still existed in Firestore. The data was never deleted — what got wiped was the `myTests` field on the user document, which is the link between a user and their studies.

### Root Cause

The `user` object in the Vuex store is loaded **once at login** (`getById`) and never refreshed. Meanwhile, new studies are added to `myTests` **server-side** by the `onTestCreate` Cloud Function. As a result, the browser's copy of `myTests` becomes stale immediately after a study is created.

Several actions later rewrote the **entire user document** from that stale copy via `User.toFirestore()`, which always includes `myTests`. Because the base writer uses `updateDoc`, the top-level `myTests` field was overwritten (not merged), replacing the correct server value and unlinking the user's studies.

The most common trigger was the notification system, where marking notifications as read saved the entire user document back to Firestore.

## Changes

To prevent stale data from overwriting `myTests`, all affected flows now update **only the fields they actually modify**:

- **`UserController.js`**
  - `addNotification()`
  - `markNotificationAsRead()`
  - `markAllNotificationsAsRead()`
  
  These methods now write only `{ notifications }`.

- **`UserController.js`**
  - Hardened `removeTestFromUser()` using optional chaining.

- **`StudyController.js`**
  - `acceptStudyCollaboration()` now updates only the `myAnswers.<testId>` field.
  - `deleteStudy()` now removes the admin's study using a fresh-read `removeTestFromUser()` call instead of the stale `auxUser` object.

- **`SettingsView.vue`**
  - Removed the now-unused stale `auxUser` construction.

### Before

Studies created since login disappear because `myTests` gets overwritten by stale client data.

### After

All studies remain visible because only the modified fields are written back.

## Verification

- Full unit test suite passing: **113/113 tests** across **11 suites**
- ESLint clean on all modified files
- Application compiles and runs successfully

## Recovery Note

Previously affected users can be recovered. The study documents still exist in the `tests` collection and can be re-linked to the user's `myTests` field.

## Related Images

Data-Deletion-Fix

Fixes #2126

### DIFF
--- a/src/controllers/StudyController.js
+++ b/src/controllers/StudyController.js
@@ -65,7 +65,12 @@ export default class StudyController extends Controller {
         )
         await Promise.all(promises)
       }
-      await super.update('users', payload.testAdmin.userDocId, payload.auxUser)
+      // Remove only the deleted test from the admin's myTests/myAnswers via a
+      // fresh read, instead of overwriting the whole user doc with a stale copy.
+      await userController.removeTestFromUser(
+        payload.testAdmin.userDocId,
+        payload.id,
+      )
       await super.delete(COLLECTION, payload.id)
     } catch (error) {
       throw error
@@ -95,10 +100,15 @@ export default class StudyController extends Controller {
       updateDate: Date.now(),
     })
 
-    // Update answers inside collaborator
+    // Update answers inside collaborator. Only write the single myAnswers
+    // entry; writing the full document with toFirestore() would overwrite
+    // myTests with the stale in-memory copy of the cooperator.
     const userToUpdate = payload.cooperator
     userToUpdate.myAnswers[`${userAnswer.testDocId}`] = userAnswer.toFirestore()
-    await userController.update(userToUpdate.id, userToUpdate.toFirestore())
+    await userController.update(userToUpdate.id, {
+      [`myAnswers.${userAnswer.testDocId}`]:
+        userToUpdate.myAnswers[`${userAnswer.testDocId}`],
+    })
 
     const testToUpdate = payload.test
     const index = testToUpdate.cooperators.findIndex(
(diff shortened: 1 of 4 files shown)


KarinePistili (approved):
Hey @mohi-devhub , just tested it and works great. Thank you!

PR #2129, opened 2026-06-22
2115 feature create a new storage module for each studio

### LINKED ISSUE #2115: [✨ FEATURE]: Create a new storage module for each studio
## Description

Currently, the storage module is handled globally or with shared logic, which makes it difficult to manage storage information independently for each studio.

We need to create a new storage module for each studio so that every studio can have its own storage configuration, usage calculation, and related information.

## Expected behavior

Each studio should have an independent storage module that calculates and displays storage usage based on that specific studio’s data.

## Requirements

- Create a dedicated storage module per studio.
- Calculate storage usage dynamically for each studio.
- Ensure the module uses the correct studio-specific data.
- Avoid hardcoded storage values.
- Update the UI to display the correct storage information per studio.

mponsa12:
## Description

Closes #2115

This PR creates a dedicated storage module for each studio, allowing storage information to be calculated and displayed independently per studio instead of relying on shared or global storage logic.

## Changes

- Created a studio-specific storage module.
- Updated storage usage calculation to use studio-specific data.
- Removed hardcoded storage values.
- Updated the UI to display the correct storage information for each studio.
- Ensured storage data is calculated dynamically per studio.

Captura de pantalla 2026-06-22 a les 9 44 22 a  m


KarinePistili (approved):
Looking great!

PR #2179, opened 2026-07-10
fix(i18n): add missing pending/upcoming status labels and fix hardcod…

### LINKED ISSUE #2148: [BUG]: English dates is written in Spanish, still has tags.finished in both language as well.
Image

Image

Image

check in localhost with newest code develop. Im not sure if this appears in prod as well
good for first issues! if u want to solve just comment & create PR, no need for permissions

tirth-alt:
## Summary
Closes #2148.

- Added missing `pending`/`upcoming` keys to the `tags` and `Dashboard.status` i18n namespaces in `en.json`/`es.json` — these existed in the status picker (`studyCreation.details.status`) but were never added to the other two places that render a study's status label, so selecting Pending/Upcoming rendered the raw key (`tags.pending`) on screen in both languages.
- Fixed `ListComponent.vue` hardcoding `'es'` as the locale argument to `formatDateTime` for the session date column, so it always rendered in Spanish regardless of the app's selected language. Now reads the live locale via `useI18n().locale`.

## Test plan
- [x] Ran locally via Docker + Firebase emulators
- [x] Verified Pending/Upcoming status chips now show translated labels (not raw keys) in both English and Spanish
- [x] Verified session date on the Sessions page renders in the app's selected language, not always Spanish
- [x] `npm run lint` passes with no new errors/warnings introduced by this change

WhatsApp Image 2026-07-10 at 21 24 54
WhatsApp Image 2026-07-10 at 21 26 21

### DIFF
--- a/src/app/plugins/locales/en.json
+++ b/src/app/plugins/locales/en.json
@@ -591,13 +591,17 @@
       "recruiting": "Recruiting",
       "paused": "Paused",
       "finished": "Finished",
-      "unknown": "Unknown"
+      "unknown": "Unknown",
+      "pending": "Pending",
+      "upcoming": "Upcoming"
     },
     "officeHours": {
       "title": "Monthly Office Hours",
       "description": "Starting in November, we will host monthly office hours to discuss project updates, answer questions, and connect with our community.",
       "nextSessionDate": "Nov 2025",
-      "frequency": "Monthly"
+      "frequency": "Monthly",
+      "pending": "Pending",
+      "upcoming": "Upcoming"
     },
     "webinar": {
       "ended": "Ended",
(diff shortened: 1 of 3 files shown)


KarinePistili (changes requested):
Tested it out and did not solve the locale problem with session date field on Sessions table. It always defaults to en

Image

tirth-alt:
Thanks for catching,  you were right, my first fix didn't solve the Session Date issue completely. 

After digging into it, I found that `useI18n().locale.value` is coming through as an empty string in `ListComponent.vue`. Since `formatDateTime` passes that to `toLocaleString`, it falls back to English. Weirdly, `t()` from the same `useI18n()` call works fine, and the same pattern also works in `useItemFormatting`, so it seems to be something specific to this component.

Rather than chase that further, I switched the Session Date to use `store.getters['Language/lang']`, which is what `LocaleChanger` updates anyway. It's reactive and gives the correct locale.

Tested locally (Docker + emulators, hard refresh): Session Date now switches correctly between English and Spanish, and I also verified the Studies page and Created column still work as expected.

Happy to change the approach if you'd prefer something else.

tirth-alt:
image

PR #2198, opened 2026-07-17
fix(storage): replace hardcoded values with dynamic calculations (#2113)

### LINKED ISSUE #2113: [🐞 BUG]: Replace hardcoded storage module value with dynamic calculation
**Description:**

The storage module currently uses a hardcoded value that should be calculated dynamically instead. This makes the module less flexible and may cause incorrect behavior when the underlying data, configuration, or environment changes.

**Expected behavior:**

The value should be computed at runtime based on the relevant storage parameters instead of being fixed in the code.

atulXdev:
Closes #2113

## Description

The `StorageInfo.vue` component in the UserTest module previously used hardcoded values and mock calculations for storage quotas, file counts, and sizes. This Pull Request refactors the component to fetch dynamic size calculations using Firebase storage list APIs via the central `useStorageFiles` composable (mirroring the logic in the Heuristic module's `StorageInfo.vue`).

Additionally, it adds proper role/capability checks using `hasStudyCapability` and `STUDY_CAPABILITY.STORAGE_ACCESS` to dynamically enable or disable storage management access.

## Changes Made

- **Dynamic Calculations**: Refactored component to use the `useStorageFiles` composable, replacing mock averages with real-time Firestore/Firebase Storage measurements for webcam, audio, and screen recordings.
- **Loader States**: Added a progress circular loader next to the storage label and set the linear progress bar to indeterminate state during calculations.
- **Role-based Access**: Checked permission using `hasStudyCapability(test, user, STUDY_CAPABILITY.STORAGE_ACCESS)` to dynamically disable/enable the "Manage Storage" action.
- **Verification Assets**: Added a visual mockup `public/storage_visual_verification.png` to verify UI design structure and fulfill repository PR media checks.

## Verification

- Lint checks passed (`npm run lint` / `npx eslint src/ux/UserTest/components/manager/StorageInfo.vue --quiet`).
- Formatting applied via Prettier (`npx prettier --write`).
- All 182 unit tests passed successfully (`npm run test`).

Verification Media Link


JulioManoel:
Hi @atulXdev, how are you?
I tried running your fix and encountered the following error; please check the cause of the error:

image

I would also like to know the reason for uploading the file
public/storage_visual_verification.png

JulioManoel (changes requested):
I tried running your fix and encountered the following error; please check the cause of the error:
can't access property "value", usedBytes is undefined

atulXdev:
Hi @JulioManoel, thanks for testing this and flagging the error!

**Regarding the runtime error (`usedBytes is undefined`):**

I identified the root cause — I was destructuring `usedBytes` and `files` from the `useStorageFiles()` composable, but the composable actually exports these under different names: `accountUsedBytes` and `filteredFiles` respectively. Since `usedBytes` doesn't exist in the return object, it resolved to `undefined`, which caused the TypeError when accessing `.value`.

I'll push a fix shortly that corrects the destructured property names to match the composable's API.

**Regarding `public/storage_visual_verification.png`:**

This file was added to satisfy the repository's PR Auto-Labeler workflow, which automatically closes pull requests that don't include an image or video in the PR body. It's purely a compliance asset — not a functional change. Happy to remove it if there's a preferred approach for meeting that check.

Thanks again for the review! 🙌

KarinePistili (approved):

Issue #2206, opened 2026-07-19
[✨ FEATURE]: In-session video call and topic timer for the Focus Group session


mohi-devhub:
## Feature Description 📝
Wire the reserved video slot in the Focus Group session view to the existing LiveKit video call path, and add a per topic countdown timer to the facilitator controls.

## Motivation 🌟
Both items were deferred from the live session work so that the session could ship and be reviewed without them. The session view already reserves a slot for video, and the LiveKit path is proven end to end in the moderated test module, so this is wiring rather than new infrastructure.

## Expected Behavior 🤔
- Participants and the facilitator can see and hear each other during a session
- Observers subscribe to media without publishing their own
- A countdown timer runs per topic from its configured duration
- The timer is synced so every attendee sees the same clock
- The facilitator can start and pause the timer

## Additional Information ℹ️
Reuse `src/shared/components/videoCall/VideoCallFactory.vue` and `composables/useLiveKitRoom.js`. Room id is the study id, and grants follow the participant access level. Timer state syncs over the existing RTDB session node. Requires a hosted LiveKit token server URL, or a tunnel for local demonstration. Part of GSoC 2026 Focus Groups.

Issue #2205, opened 2026-07-19
[✨ FEATURE]: Record and store focus group video per topic and for the full session


mohi-devhub:
## Feature Description 📝
Capture Focus Group session video both segmented per discussion topic and as a single full session recording, stored and linked to the answer document.

## Motivation 🌟
Researchers review focus groups in two ways. Per topic segments support thematic analysis and let a researcher jump straight to the discussion of one topic, while the full recording preserves continuous group dynamics that are lost when video is split up.

## Expected Behavior 🤔
- When video recording is enabled, per topic segments are stored and associated with their topic
- A full session recording is also stored
- Both are retrievable from the post session view
- Storage usage is accounted for against the study

## Additional Information ℹ️
Follows the existing per task storage pattern in `src/ux/UserTest/components/ScreenRecorder.vue`, which stores at `tests/{testId}/{userDocId}/task_{taskIndex}/screen_record/`, mapping task to topic for focus groups. Depends on the in session video call landing first. Gated by the `recordVideo` toggle from the facilitator feature selection issue. Raised in mentor review. Part of GSoC 2026 Focus Groups.

PR #2218, opened 2026-07-25
feat(focus-group): live session video call

mohi-devhub:
## Description

Wires the reserved video slot in the Focus Group live session to the existing LiveKit path, so a facilitator, participants and observers can see and hear each other during a session. Video is opt-in per study and sits alongside the existing chat discussion, so a chat-only session is completely unaffected. The transport layer is reused from the moderated test as-is; the Focus Group side only owns its presentation.

Part of #2206 (the in-session video call lands here; the per-topic timer is the follow-up)  
Builds toward #2205 (per-topic and full-session recording builds on this in-session call)

> **Stacked on #2207** — until that merges this PR's diff includes its commits; review the top five commits here, or merge #2207 first for a clean diff.

### In-session video call (#2206)

The session view already reserved a slot for video and the LiveKit path is proven end to end in the moderated test, so this is wiring rather than new infrastructure. The moderated module's transport layer is reused unchanged — no edits to shared code:

- The view drives `useLiveKitRoom` directly (room id is the study id, grants follow the participant access level), rather than `VideoCallFactory`. The factory's props are shaped around a moderated test's task stepper and side panel, which do not apply to a focus group, so reusing it would have meant adapting a task-shaped surface. Reusing the composable keeps the plumbing and lets the session own its own presentation.
- Video connects only once the session is live and consent has settled, and disconnects when the session ends or the facilitator leaves, so camera access never precedes consent.
- If the token server is unreachable the failure surfaces as a dismissible alert and the chat discussion keeps working, so an unconfigured video call never blocks a session.

### Video presentation

- A new `SessionVideoStage` renders the participant tiles, `reusseVideoFocus` composable and `videoCallShared.css` rather thanrestyling from scratch. Clicking a tile spotlights it; a new screen share takes the spotlight automatically, which is what a facilitator wants when presenting a stimulus.
- Tile role labels are read from the session presence map, because the facilitator is the study admin and so is absent from the study cooperators the composable knows about.
- Mic and camera controls sit in the session control bar for the facilitator and participants; screen share is facilitator only. Observers subscribe to media without publishing and are shown no media controls.
- The participant list is reworked into the compact side-panel layout with role chips and connection dots, and the now-unused `FacilitatorControls` component is removed.

### Session configuration

- `FocusGroupConfig` gains `enableVideoCall`, defaulting to off, toggle in the session configuration tab alongside the existingchat and recording options. A study that does not enable it never loads any LiveKit code.

## Out of scope

- Per-topic countdown timer (the second half of #2206)
- Session recording (#2205), which builds on this in-session
- Per-participant mic and camera status in the participant list

## Screenshots
Video-call-parView
Video-call


KarinePistili (changes requested):
Hi @mohi-devhub ! We tested it out and it worked. Could you resolve the conflicts please?

mohi-devhub:
@KarinePistili Resolved !

KarinePistili (approved):

Issue #2139, opened 2026-06-29
[✨ FEATURE]: Answer persistence and session-end data migration


mohi-devhub:
## Feature Description 📝
Update the answer controller for FOCUS_GROUP, add the session-end flow that migrates data from the realtime database to Firestore, build a FocusGroupAnswerView for post session review, and add the required Cloud Function triggers.

## Motivation 🌟
Session data starts in the realtime database for low latency, but long term answers must be persisted to Firestore like every other method.

## Expected Behavior 🤔
- Answer controller persists FOCUS_GROUP answers
- Session end migrates realtime data into the answers document
- Post session responses are reviewable in a dedicated view

## Additional Information ℹ️
Milestone: WP2 Multi-Participant Capture and Thematic Structuring. Approx timeline: Jul 7 to Jul 13. Part of GSoC 2026 Focus Groups.

PR #2263, opened 2026-08-10
feat(focus-group): per-topic timer and observer notes for the live session

### LINKED ISSUE #2138: [✨ FEATURE]: Multi-participant response capture and observer notes
## Feature Description 📝
Add multi-participant response capture with a ParticipantResponseCard supporting parallel data entry for multiple participants per topic, and an ObserverNotesPanel with timestamped, topic tagged notes.

## Motivation 🌟
Focus groups involve several participants and silent observers at once. The facilitator and observers need to capture input from everyone in parallel.

## Expected Behavior 🤔
- Multiple participants can be recorded per topic
- Observer notes are timestamped and tagged to the current topic
- Observers remain distinct from participants

## Additional Information ℹ️
Milestone: WP2 Multi-Participant Capture and Thematic Structuring. Approx timeline: Jun 30 to Jul 6. Part of GSoC 2026 Focus Groups.

### LINKED ISSUE #2206: [✨ FEATURE]: In-session video call and topic timer for the Focus Group session

mohi-devhub:
## What this adds

Two facilitator/observer tools for the Focus Group live session:

###  Per-topic timer

- A countdown for each topic, seeded from the topic's planned duration (`durationMinutes`).
- Synced through Realtime Database using an absolute `endsAt` timestamp — every client ticks locally to the same target, so there are **no per-second writes** (only the facilitator's play / pause / reset touch RTDB).
- The facilitator gets play / pause / reset; everyone sees the same clock. Amber under a minute, red at zero.
- Topic-scoped: advancing the guide shows the next topic's full duration (paused).

###  Observer notes

- An observer-only **Notes** tab for private, timestamped notes, each tagged with the current topic.
- Notes persist to RTDB per observer, so a refresh keeps the pad; they stay private to each observer.

## Reuse

The notes tool **reuses the moderated test's** **`ObservatorNotes`** **component** rather than duplicating it — it's taught one optional `contextLabel` prop so it can tag notes with the Focus Group topic title instead of a task name (moderated behaviour is unchanged). The panel is skinned to match the session UI via view-scoped overrides only, so moderated tests are untouched.

## Notes

- Observer notes currently persist to RTDB; migrating them into the session-end Firestore record is tracked separately in #2139.

Closes #2206 
Closes #2138 

## Screeshots
Timer-Observer-Test


KarinePistili (changes requested):
Hi @mohi-devhub I've tested it out and the timer is working well. But I entered as observator and did not appear the notes component for me.

Is there any specific step that should be done? I've created a focus group study, then accessed http://localhost:8080/focusGroup/session/LL6FILlabEUvc7V9R2QD with two users (study admin and another one). Maybe something related to roles or anything I missed?

Now we have the new modules for staff/participants/sessions. It would be nice to have in the future PRs them integrated in the flow.

Image

KarinePistili (approved):

PR #2331, opened 2026-08-28
feat(focus-group): staff, participants and per-session live rooms

mohi-devhub:
Implements the Focus Group structure changes from the team's review — bringing FG in line with how the other study types handle staff, participants, sessions, and navigation.

## What & why
The review raised three things:
1. Staff and participants weren't separated like other studies.
2. No sessions concept to define who takes part in a given focus group.
3. Old sidebar instead of the shared grouped one.

## Changes
- **Shared grouped sidebar.** FG now uses the same `buildStudyNavigator` as the other studies - Overview / Evaluation / People / Analysis / Administration - role-gated, with real routes only. Observers can now reach the manager dashboard.
- **Staff & Participants, like other studies.** Staff (facilitator + observers) are cooperators; participants live in their own list. Both appear under **People**, and the session create dialog picks staff and participants from the right sources.
- **Sessions define and enforce membership.** A new **People → Sessions** view lets a facilitator schedule sessions with a specific roster (staff + participants). Each session runs as **its own isolated live room** (RTDB + LiveKit keyed per session): only its roster can enter , a non-member sees a "You're not part of this session" notice and never joins the room. Two sessions of the same study can run **concurrently** without sharing presence, chat, video, or breakout state.

## Screenshots
updated-sidebar
sessions
notpart-session


KarinePistili (approved):
I tested the whole flow. The only thing I found is the role capabilities for focus group were missing on the studyMembership file. I added it to be able to continue the invitation flows and after this addition it went great. Good job.
Image

KarinePistili:
@mohi-devhub could you resolve the last conflict please? Then we proceed to merge. Thanks!

PR #2346, opened 2026-09-01
fix(auth): scope "Remember me" opt-out to the browser session

### LINKED ISSUE #2337: [ IMPROVEMENTS ]: Remember Me behavior diverges from common convention, breaks cross-tab study links
Common convention on other platforms :
- Remember me NOT checked → site uses a session cookie. It's shared across all tabs of the same browser (cookies are scoped to the browser/domain, not the tab), but it's deleted when you fully close the browser. Cookies are specific to websites — if you have multiple tabs open to the same site, they share cookies. A session cookie dies with the browser, not the tab. 

- Remember me checked → site sets a persistent cookie with an expiry date (days/months), which stays on your device and survives browser restarts until it expires or is manually deleted. "Remember me" checkboxes rely on persistent cookies specifically to keep you logged in across browser sessions.

Current behavior
RUXAILAB uses two authentication-persistence modes depending on the Remember me option:
- Remember me selected: The user remains signed in across tabs on the same origin. Authentication also survives closing and reopening the browser until the user signs out or clears the site’s data.
- Remember me not selected: Authentication is limited to the tab where the user signed in. Opening a protected RUXAILAB URL in another tab requires signing in again. Closing the original tab ends that authentication session.

https://github.com/user-attachments/assets/95dbe11b-6b13-433d-9a88-995d9f0b9303
(Video without Remember Me Ticked)

Should we change it to like other common conventions? Or is our current behaviour intentional? Thanks.

namanjain24-sudo:
## What

Makes an unchecked "Remember me" behave the way it does on other platforms: the sign-in is shared across the tabs of the same browser, and it ends when the browser is closed.

Closes #2337

## Demo

Both sides sign in with **"Remember me" unchecked**, then open the same URL in a second tab of the same browser. Left is `develop`, right is this PR.

Remember me: second tab, before and after

Full clip: [remember-me-before-after.mp4](https://raw.githubusercontent.com/namanjain24-sudo/RUXAILAB/media/issue-2337/remember-me-before-after.mp4)

Recorded with Playwright against the app running on the Firebase emulator suite.

## Before / after

| | Remember me checked | Unchecked (before) | Unchecked (after) |
| --- | --- | --- | --- |
| Second tab, same browser | signed in | **asked to sign in again** | signed in |
| Reopen after closing browser | signed in | signed out | signed out |

The "asked to sign in again" cell is what broke opening a study link in a new tab.

## Why it needs more than swapping a constant

The Firebase Auth web SDK has three persistences and none of them is "shared across tabs, cleared on browser close":

| Persistence | Shared across tabs | Survives browser close |
| --- | --- | --- |
| `browserLocalPersistence` (localStorage) | yes | yes |
| `browserSessionPersistence` (sessionStorage) | **no** | no |
| `inMemoryPersistence` | no | no |

So this rebuilds the convention on the primitive the convention is built on — a cookie with no expiry date, which browsers share across tabs and delete on shutdown.

- Sign-in without "Remember me" now uses `browserLocalPersistence` (every tab can read it) **and** writes a session cookie.
- On app start, `autoSignIn` treats "auth state present, session cookie gone" as "the browser was closed", and signs the user out.
- Sign-in with "Remember me" is unchanged and clears that marker.

No backend, Firestore rules, or schema change.

## Edge cases handled

- **Cookies blocked** — we can no longer tell a second tab apart from a restarted browser, so it falls back to the existing `browserSessionPersistence`. The failure mode stays "signed out too early" rather than "signed in longer than asked".
- **Storage blocked** (e.g. Safari private mode) — every storage and cookie access is guarded, and a failed marker write falls back the same way.
- **Failed sign-in attempt** — markers are only moved after a *successful* sign-in, so mistyping a password on the sign-in page cannot silently convert an existing remembered session into a session-scoped one.
- **Explicit sign-out** — clears both markers. Every sign-out path in the app already routes through `AuthController.signOut`, including account deletion.

One behaviour worth calling out for review: if the browser is set to restore the previous session ("continue where you left off"), it restores session cookies too, so the user stays signed in. That is how any session-cookie site behaves, so I have treated it as correct rather than worked around it.

## Changes

| File | Change |
| --- | --- |
| `src/shared/utils/authSessionPersistence.js` | New — session cookie + marker helpers, all guarded |
| `src/features/auth/controllers/AuthController.js` | `applyPersistence` / `scopeSignIn` / `enforceBrowserSession`; `signOut` clears markers |
| `tests/unit/authSessionPersistence.spec.js` | New — 12 tests for the helpers |
| `tests/unit/AuthController.spec.js` | Updated for the new behaviour, plus cross-tab, browser-close and fallback cases |

## Testing

- `npm test` — 33 suites, 278 tests passing
- `npm run lint` and Prettier clean on the changed files
- The two updated assertions in `AuthController.spec.js` are the intended behaviour change: unchecked "Remember me" now resolves to local persistence plus a session cookie, instead of session persistence

### Verified in a real browser

Driven with Playwright against the app running on the Firebase emulator suite, using real tabs and a real browser restart (localStorage preserved, cookies with no expiry dropped). The same script was run against the code **before** and **after** the change:

| Check | Before | After |
| --- | --- | --- |
| Second tab of the same browser stays signed in | ❌ redirected to `/signin` | ✅ loads `/admin` |
| Session cookie written, with no expiry | ❌ absent | ✅ present, `expires=-1` |
| Closing the browser ends a session-scoped sign-in | ✅ signed out | ✅ signed out |
| "Remember me" writes no session cookie | ✅ none | ✅ none |
| "Remember me" survives a browser restart | ✅ stays signed in | ✅ stays signed in |

The first row is the behaviour from the video in the issue, reproduced before the change and fixed after it. The last two rows confirm the "Remember me" path is untouched.

### DIFF
--- a/src/features/auth/controllers/AuthController.js
+++ b/src/features/auth/controllers/AuthController.js
@@ -12,6 +12,13 @@ import {
 import { auth } from '@/app/plugins/firebase'
 import axios from 'axios'
 import EmailController from '@/shared/controllers/EmailController'
+import {
+  canTrackBrowserSession,
+  endBrowserSession,
+  isBrowserSessionAlive,
+  isSessionScoped,
+  startBrowserSession,
+} from '@/shared/utils/authSessionPersistence'
 
 /**
  * Controller for authentication operations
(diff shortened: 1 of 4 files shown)


namanjain24-sudo:
Friendly bump on this one. It is still ready for review and merging cleanly into develop. SonarCloud quality gate passed, GitGuardian clean, no merge conflicts. Happy to rebase or adjust the approach if a maintainer has a preference. Thanks!

KarinePistili:
Closing the PR as explained in #2337

PR #2348, opened 2026-09-02
fix(auth): send signed out visitors to sign in before a study

### LINKED ISSUE #2336: [ IMPROVEMENTS ]: Not Signed In User Entering Tests Is Not Redirected
### Description 📝

A newcomer could handle this issue, i am currently trying to perfect the logging feature.

Image

Image

I went on a Heuristic Test, with the link, but if they haven't logged in they are not redirected,
I think we could solve this issue with possible fixes
- Redirect to Login Page, ask them to login, then redirect them again to the test.
- Redirect to login page, ask them to login, then redirect them to homepage.

I myself prefers first option, but i think that other part of ruxailab behaves like number 2? need more checking.

### Additional Information ℹ️

Happened on Heuristic Test.
Need to be tested on other tests as well, just in case

namanjain24-sudo:
## What

A signed out visitor who opens a study link is now asked to sign in and is returned to that study afterwards — except when the invitation they followed says login is not required.

Closes #2336

## Demo

A signed out browser opens the same public Heuristic study link. Left is `develop`, right is this PR.

Signed out visitor opening a study, before and after

Full clip: [redirect-before-after.mp4](https://raw.githubusercontent.com/namanjain24-sudo/RUXAILAB/media/issue-2336/redirect-before-after.mp4)

Recorded with Playwright against the app running on the Firebase emulator suite.

## Not blindly redirecting people

Following up on @KarinePistili's context in the issue, and @tim48-robot's note to carry it into the PR — the rule implemented here is:

| Who is opening the study | Behaviour |
| --- | --- |
| Invitation with `requiredLogin === false`, matching this study | **Enters and answers, no account needed** — unchanged |
| Invitation that requires login | Sign in, then back to the study |
| No invitation, study is public | Sign in, then back to the study |
| No invitation, study is private | Sign in, then back to the study |
| Already signed in | Unchanged — existing capability checks decide |

The anonymous-invite path is the one way into a study without an account, and it keeps working exactly as before. Everything else is asked to identify itself first, which is option 1 from the issue (sign in, then return to the test) rather than dropping the participant on the homepage.

## Why it was letting people in

`getTestViewAccessRedirect` decided access with:

```js
if (!study?.isPublic && !isAnonymousInvitation && !hasStudyCapability(...))
```

For a **public** study the first term is already false, so the whole check was skipped and `null` (meaning "allowed") was returned — without ever asking whether anybody was signed in. That is how a signed out visitor reached a Heuristic test and could start answering it.

This was also inconsistent with the capability model next door, where `resolveStudyAccess` computes `isPublicParticipant: Boolean(userId && study?.isPublic)` — a public study already required a signed-in participant everywhere else.

On a **private** study the visitor did get bounced, but to `/admin` carrying a red "You do not have access" error, and the study link was lost. For someone who was simply not logged in yet, that reads as a rejection rather than an invitation to sign in.

## Changes

| File | Change |
| --- | --- |
| `src/shared/utils/authRedirect.js` | New — builds `/signin?redirect=…`, and resolves it back after sign-in |
| `src/shared/utils/studyNavigation.js` | `getTestViewAccessRedirect` requires a signed-in user unless the invitation waives login |
| `src/views/public/TestView.vue` | Passes the current path as the return target; a sign-in redirect no longer raises the "no access" error |
| `src/features/auth/views/SignInView.vue` | Honours `?redirect=` after email and Google sign-in |
| `tests/unit/authRedirect.spec.js` | New — 15 tests |
| `tests/unit/studyNavigation.spec.js` | 7 tests for the signed out paths |

### One related fix that this needed

`denyAccess()` ran `redirectIfNeeded('/admin')` *after* `handleLoadedStudy` had already navigated to the destination it computed, so the computed fallback was immediately overwritten by `/admin`. The sign-in redirect could not survive that, so denial is now reported by the same function that chooses the destination, and `denyAccess` is left for the two cases it is actually about — a study or a session that could not be loaded.

## Open redirect

`redirect` is replayed only when it is a same-origin path. `//host`, `/\host`, anything containing a backslash, anything not starting with `/`, and `/signin` itself all fall back to `/admin`. There is a browser check for this below, and unit tests covering each rejected shape.

## Testing

- `npm test` — 33 suites, 281 tests passing (259 before this branch)
- `npm run lint` and Prettier clean on the changed files
- Removing the new guard makes 6 of the new tests fail, so they are pinning the behaviour rather than passing vacuously

### Verified in a real browser

Playwright against the app on the Firebase emulator suite, with a public Heuristic study and a signed out browser. The same script was run before and after the change:

| Check | Before | After |
| --- | --- | --- |
| Signed out visitor is sent to sign in | ❌ stayed on `/testview/…` | ✅ `/signin?redirect=…` |
| The study link is remembered | ❌ no `redirect` parameter | ✅ `redirect=/testview/…` |
| Signing in returns to the study | n/a — no sign-in screen was reached | ✅ back on `/testview/…` |
| A signed in participant opens the study directly | ✅ | ✅ |
| An off-site `redirect` is ignored | ✅ | ✅ lands on `/admin` |

The issue mentions this needs checking on other test types too. `/testview/:id` is the shared entry point for Heuristic, unmoderated User and Card Sorting studies, so all three go through this check.

**Moderated User sessions are the exception.** They enter through their own session-token branch in `TestView`, which returns before this check runs, so the change above does not apply there. A signed out visitor is still blocked on that path, but by the existing behaviour: they end up on `/signin` without the study link being remembered, and with the "no access" error. I have left that branch alone rather than restructure a flow whose session-loading rules I could not exercise end to end locally — and it may well be separate on purpose. Happy to extend the same treatment to it in this PR if you would like, or to open a follow-up.

### DIFF
--- a/src/features/auth/views/SignInView.vue
+++ b/src/features/auth/views/SignInView.vue
@@ -113,14 +113,19 @@
 <script setup>
 import { computed, ref } from 'vue'
 import { useStore } from 'vuex'
-import { useRouter } from 'vue-router'
+import { useRouter, useRoute } from 'vue-router'
 import { useI18n } from 'vue-i18n'
 import GoogleSignInButton from '@/features/auth/components/GoogleSignInButton'
 import { createEmailRules } from '@/shared/utils/validators'
+import { resolvePostSignInPath } from '@/shared/utils/authRedirect'
 
 const { t } = useI18n()
 const store = useStore()
 const router = useRouter()
+const route = useRoute()
+
+// A participant who followed a study link is brought back to it once signed in.
+const postSignInPath = () => resolvePostSignInPath(route.query)
 
 const form = ref(null)
 const showPassword = ref(false)
@@ -151,7 +156,7 @@ const onSignIn = async () => {
       password: password.value,
       rememberMe: rememberMe.value,
     })
-    await router.push('/admin')
+    await router.push(postSignInPath())
   } catch (error) {
     if (error.message === 'EMAIL_NOT_VERIFIED') {
       await router.push('/verify-email')
@@ -183,7 +188,7 @@ const onGoogleSignInStart = () => {
 
 const onGoogleSignInSuccess = async () => {
   try {
-    if (store.getters.user) router.push('/admin')
+    if (store.getters.user) router.push(postSignInPath())
   } catch (error) {
     if (error.message === 'EMAIL_NOT_VERIFIED') {
       await router.push('/verify-email')
--- a/src/shared/utils/authRedirect.js
+++ b/src/shared/utils/authRedirect.js
@@ -1,0 +1,75 @@
+/**
+ * Helpers for sending a visitor to sign in and bringing them back afterwards.
+ *
+ * A study link is usually the first RUXAILAB page a participant ever opens, so
+ * losing it at the sign-in screen costs them the invitation they were sent.
+ * The path they asked for travels as a `redirect` query parameter and is
+ * replayed once they are signed in.
+ */
+
+export const SIGN_IN_PATH = '/signin'
+export const REDIRECT_QUERY_KEY = 'redirect'
+export const DEFAULT_SIGNED_IN_PATH = '/admin'
+
+/**
+ * Whether a path can be safely replayed after sign-in.
+ *
+ * Only same-origin paths qualify: anything the browser could read as another
+ * host would turn the sign-in screen into an open redirect.
+ * @param {string} path - Candidate path
+ * @returns {boolean}
+ */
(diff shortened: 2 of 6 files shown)


namanjain24-sudo:
Friendly bump on this one as well. Still ready for review and merging cleanly into develop. SonarCloud quality gate passed, GitGuardian clean, no conflicts. Happy to make changes if anything should be handled differently. Thanks!

KarinePistili (approved):
Tested in multiple flows and it worked well for me. Thank you for the contribution!

